PR #19, opened 2016-09-27
Make NiceMock central, so all created mock objects are NiceMocks

LukasDistel:


MathMagique:
@LukasDistel Excellent! Thanks a lot!

Issue #24, opened 2016-12-04
Apache Arrow result sets


xhochy:
In line with `turbodbc_numpy` it would be helpful to also have a `turbodbc_arrow` library that will produce Apache Arrow Tables / RecordBatches.

MathMagique:
That seems very promising, in particular to open up turbodbc for many tools that already rely or plan to use arrow.

yaxxie:
When this gets released, will we bump to 1.2.0?

MathMagique:
I plan to bump to 2.0.0 and adpot semantic versioning from then on. I will also make a breaking change for the deprecation I introduced with `make_options()`.

yaxxie:
Even better 👍 Thanks!

PR #26, opened 2016-12-12
Add Apache Arrow result set support as turbodbc_arrow

xhochy:


xhochy:
@MathMagique I've added in the latest commit C-only (i.e. boost-free) routines to translate SQL Datetimes/Dates to ms/us. Might be interesting for other parts of `turbodbc`.

MathMagique:
Those routines sure are interesting - another step to get rid of boost :-)

xhochy:
@MathMagique one problem I face here is that `turbodbc_arrow_unit_test` depends on linking against `libpython`. I suspect that this is not available on all platforms. Any suggestions on how to deal with that?

MathMagique:
The Python library is not necessarily called "libpython" on each system. Pybind11's cmake scripts will expose a `${PYTHON_LIBRARY}` variable you should be able to use:
https://github.com/pybind/pybind11/blob/f3de2d5521c3e95e3212b7c5dc26d9d4e23b17fb/tools/FindPythonLibsNew.cmake#L166

xhochy:
Either we go ahead and use `conda` in the Travis setup to install Arrow or this PR will be blocked at the moment by https://issues.apache.org/jira/browse/ARROW-566. It is definitely possible to build wheels that satisfy the requirements for `pip` based `turbodbc_arrow` but not as simple as with `conda`.

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	int64_t date_to_days(char const * data_pointer)
+	{
+		auto & sql_date = *reinterpret_cast<SQL_DATE_STRUCT const *>(data_pointer);
+
+    struct tm date = {0};
+    date.tm_year = sql_date.year - 1900;
+    date.tm_mon = sql_date.month - 1;
+    date.tm_mday = sql_date.day;
```
yaxxie: Should this not also have -1? since the definition for std::tm is "since 1900 01 01"?

yaxxie: I understand why this is wrong now.

yalwan-scee:
Just out of curiosity, does this implement a fetchbatch or fetchmany method?

xhochy:
At the moment it only implements a fetchall method but adding fetchbatch/fetchmany wouldn't be that hard. First I would like to get the branch merged :)

yalwan-scee:
Before it diverges too much ;)

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -81,8 +75,8 @@ namespace {
     epoch.tm_year = 70;
     epoch.tm_mday = 1;
 
-    // Milliseconds since the epoch
-    return lrint(difftime(mktime(&date), mktime(&epoch)) * 1000);
+    // days since the epoch
+    return lrint(difftime(mktime(&date), mktime(&epoch)) / 86400);
```
wesm: Take note of what was required for msvc compatibility https://github.com/apache/arrow/blob/master/cpp/src/arrow/python/util/datetime.h#L35

MathMagique: @wesm The master branch already supports MSVC packaging (Python 3.5+) and @xhochy has created turbodbc, so I hope there is not too much to do on the packaging side.

yaxxie:
I didn't take a look at the code, but are you able to comment on whether the datatypes coming out will be closely corresponding with those fed in? For e.g. in turbodbc_numpy all ints are coerced to int8s, which is extremely space inefficient for a series of byteints.

wesm:
Is byteint different from int8?

yaxxie:
My bad, I meant int64

wesm:
We should use the exact C types for Arrow

MathMagique:
@yalwan-scee Currently, the internal API of turbodbc is only capable of reporting 64 byte integers. That is not a fundamental issue and easily fixed, however it requires extra work. I'd like to have a first increment for the arrow support before I extend anything internally.

yaxxie:
Of course, it doesn't look like it will be hard work from what I've seen and scope creep is a killer.

Is there anything I can do to help this PR move along? I'd also like to see arrow support :)

xhochy:
@MathMagique Please proceed to review. It is now in a stage where it is not performant but such that it works and can be merged.

C++ unit tests on OSX are failing due to a missing `libpython.dylib`. We either have the option to always compile a fresh python from source or skip these tests for now. As they are included in the Python unit tests, I would vote for skipping them on OSX.

yaxxie:
@xhochy Could you elaborate on your comment about "it is not performant"?

MathMagique:
@xhochy I'll try to review this over the weekend.

xhochy:
@yaxxie There are some bits like extra copies in the code at the moment that can be done in a more efficient way. While this implementation is ready to be included in the master, it should not be used for performance comparisions yet.

xhochy:
@yaxxie Also I would like make a follow-up PR soon that should be able to return Arrow dictionary arrays that would than in turn produce Pandas Categorical columns. This should show magnitudes of performance improvements in string results.

wesm:
@xhochy if we move the dictionary encoding / hashing code from parquet-cpp into arrow, then we can create a typed dictionary array builder without too much effort

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -0,0 +1,436 @@
...
+	EXPECT_CALL(rs, do_get_buffers()).WillOnce(testing::Return(expected_buffers_1)).WillOnce(testing::Return(expected_buffers_2));
+  EXPECT_CALL(rs, do_fetch_next_batch()).WillOnce(testing::Return(OUTPUT_SIZE)).WillOnce(testing::Return(OUTPUT_SIZE)).WillOnce(testing::Return(0));
+
+  turbodbc_arrow::arrow_result_set ars(rs);
+  std::shared_ptr<arrow::Table> table;
+  ASSERT_OK(ars.fetch_all_native(&table));
+  ASSERT_TRUE(expected_table->Equals(*table));
+}
+
+TEST(ArrowResultSetTest, MultipleBatchMultipleColumnResultSetConversion)
```
MathMagique: The other tests are pretty long already. This one approaches 300 lines! Please, split it up into sensible subtests with sensible names. Use helper functions to extract common setup logic where it makes sense.

python/turbodbc_test/test_has_arrow_support.py
```diff
@@ -0,0 +1,26 @@
+from mock import patch
+
+from turbodbc.cursor import _has_arrow_support
+
+import six
+import pytest
+
+# Skip all parquet tests if we can't import pyarrow.parquet
+pa = pytest.importorskip('pyarrow')
```
MathMagique: is this assignment necessary?

python/turbodbc_test/test_select_arrow.py
```diff
@@ -0,0 +1,247 @@
...
+            cursor.execute("SELECT a FROM {}".format(table_name))
+            result = cursor.fetchallarrow()
+            assert isinstance(result, pa.Table)
+            assert result.num_columns == 1
+            assert result.num_rows == 0
+
+
+@for_each_database
+@pyarrow
+def test_arrow_int_column(dsn, configuration):
```
MathMagique: These tests are _very_ similar to each other. Try to extract a common test function.

python/turbodbc_test/test_select_arrow.py
```diff
@@ -0,0 +1,247 @@
...
+            assert result.num_columns == 1
+            assert result.num_rows == 3
+            assert result.column(0).name == _fix_case(configuration, "a")
+            assert str(result.column(0).type) == "bool"
+            assert result.column(0).to_pylist() == [True, False, True]
+
+
+@for_each_database
+@pyarrow
+def test_arrow_binary_column_with_null(dsn, configuration):
```
MathMagique: binary could be confused with the `BINARY` SQL type

.travis.yml
```diff
@@ -83,7 +83,7 @@ before_install: |
   fi
 
 install:
-  - pip install numpy==1.8.0 six twine pytest-cov coveralls
+  - pip install numpy==1.10.4 pyarrow==0.4.0 six twine pytest-cov coveralls
```
xhochy: This is the minimal numpy requirement for Arrow & Python 3 due to some bugs in NumPy Macros. We can stay on a lower version for Python 2 if needed. But as e.g. conda packages are nowadays built against 1.11+, it would probably be ok to drop some older versions of NumPy now.

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -0,0 +1,252 @@
...
+			default:
+        type = std::make_shared<arrow::StringType>();
+		}
+    fields.emplace_back(std::make_shared<arrow::Field>(column_info[i].name, type, column_info[i].supports_null_values));
+	}
+  return std::make_shared<arrow::Schema>(fields);
+}
+
+Status append_to_double_builder(size_t rows_in_batch, const std::unique_ptr<ArrayBuilder>& builder, const cpp_odbc::multi_value_buffer& input_buffer, uint8_t* valid_bytes) {
+  auto typed_builder = static_cast<DoubleBuilder*>(builder.get());
```
xhochy: Only the `shared_ptr` instance is const, not the builder itself.

MathMagique: You are right.

xhochy:
@wesm yes, that's what I just started working on in Arrow

xhochy:
@MathMagique I'll fix the indentation so that it's looks okish and after this PR is merged, I'll add `clang-format` to the build toolchain so that it's easier keep this in sync in future.

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -0,0 +1,252 @@
...
+  	if (element.indicator == SQL_NULL_DATA) {
+      ARROW_RETURN_NOT_OK(typed_builder->AppendNull());
+  	} else {
+      ARROW_RETURN_NOT_OK(typed_builder->Append(element.data_pointer, element.indicator));
+  	}
+  }
+  return Status::OK();
+}
+
+Status arrow_result_set::fetch_all_native(std::shared_ptr<arrow::Table>* out)
```
xhochy: Using these return codes comes from the Google C++ Styleguide https://google.github.io/styleguide/cppguide.html#Exceptions that is used for Apache Arrow code. As the `turbodbc_arrow` C++ is mostly Arrow code, I have sticked to it here. Thus also the pointer to the `shared_ptr` as it's an output parameter.

python/turbodbc_test/test_has_arrow_support.py
xhochy: Yes, with this statement the tests are skipped if `pyarrow` is not installed, this is needed e.g. for the unit tests on windows.

MathMagique: This should work without the assignment part...

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -0,0 +1,252 @@
...
+	  		default:
+          // Strings are the only remaining type
+          ARROW_RETURN_NOT_OK(append_to_string_builder(rows_in_batch, columns[i], buffers[i].get(), valid_bytes.data()));
+          break;
+	  	}
+		}
+		rows_in_batch = base_result_.fetch_next_batch();
+	} while (rows_in_batch != 0);
+
+  // TODO(ARROW-XXX): Add convenience constructor for Table from Schema+Arrays
```
xhochy: It's already implemented in Arrow ;)

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -0,0 +1,436 @@
...
+  turbodbc_arrow::arrow_result_set ars(rs);
+  auto schema = ars.schema();
+  ASSERT_EQ(schema->num_fields(), 1);
+  auto field = schema->field(0);
+  ASSERT_EQ(field->name(), "int_column");
+  ASSERT_EQ(field->type(), arrow::int64());
+  ASSERT_EQ(field->nullable(), true);
+}
+
+TEST(ArrowResultSetTest, AllTypesSchemaConversion)
```
xhochy: I'll refactor some of the tests but this one is quite comprehensive already. Splitting it up will probably not produce any better understandable code.

MathMagique:
Quite the beast of a pull request! Excellent work @xhochy, and many thanks! I plan a release for the near future with the new arrow feature listed as being in "alpha" status. This way, we can collect feedback while users know that improvements will follow.

MathMagique:
Fixes #24

wesm:
This is great! It'd be great to get the conda-forge packages updated to make it easier for some users to kick the tires. I think this is the first usage of the pyarrow C++ API

napjon:
HI guys, could we update turbodbc in conda-forge?

PR #29, opened 2016-12-14
Extract steps in Travis into separate scripts

xhochy:
Makes building things like a Docker container with the same dependencies easier.


MathMagique:
Thanks @xhochy!

PR #36, opened 2017-01-23
CI: Remove sudo requirement

xhochy:
If this succeeds, Travis builds should be much faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 dist: trusty
-sudo: required
+sudo: false
 
 python:
   - "2.7"


xhochy:
Tests fail due to MySQL connectivity problems. It seems like the Python tests connect to MySQL without using a password but actually the DB has a password (which is simply blank).

xhochy:
See also: https://docs.travis-ci.com/user/database-setup/#MySQL

MathMagique:
Weird; the same configuration works on master with sudo. Maybe there are version differences. You could try setting a real password for the MySQL database and use it in the `odbc.ini` file.

Build times seem to have gone up by a few minutes compared to https://travis-ci.org/blue-yonder/turbodbc/builds/194303987, by the way.

xhochy:
That is kinda frustrating as Travis actually tells one that it should be the same image just without sudo, faster build starts and 0.5 CPUs additionally.

PR #61, opened 2017-03-19
README Development Version update

keitherskine:
This PR is an attempt to clarify the way that turbodbc can be built from scratch.  It fills in some gaps in the original description.


MathMagique:
Keith, that's just brilliant! Could not have written it any better myself, and that's probably why you had do it yourself in the first place ;-). Thanks a lot for your contribution.

yaxxie:
Thank you for this

PR #67, opened 2017-03-23
Removing numpy date range restrictions

yaxxie:
Using travis to check if integration tests are ok, I don't have a local setup for these


yaxxie:
It failed but looking at the failures I don't think it is to do with the commits? Is the build system broken currently?

MathMagique:
I guess the problem is this: https://travis-ci.org/blue-yonder/turbodbc/jobs/214425586#L706

It seems like lots of values such as the epoch are left uninitialized; anything can happen then at a later stage. The master build itself is working.

yalwan-scee:
Hmm, I did not get uninitialised variables warning locally, and the build for the other PR passed when I used a similar pattern

It looks to me like this is the issue:

https://travis-ci.org/blue-yonder/turbodbc/jobs/214425586#L3890

no module named pybind

MathMagique:
The "no module named pybind" is a rather unfortunate error that occurs because pip does not correctly install pybind11 when it is trying to build a wheel. `pip` will then perform another attempt to install the package, which then usually succeeds. This also happens in your case: https://travis-ci.org/blue-yonder/turbodbc/jobs/214425586#L3897

The build fails because the integration test fails, and I suppose it has something to do with the unitialized variable thingy.

MathMagique:
Also, I have reported a bug with the pybind11 creators for this: https://github.com/pybind/python_example/issues/16

MathMagique:
Do you have an explanation why my name pops up as the author in the commit you added to this pull request?

MathMagique:
Another suggestion as to why this could fail: Does numpy support dates outside of a certain range?

yalwan-scee:
Yes sorry about that, I realised you could spoof committers and picked the first E-mail I could see to verify, I forgot to change it back afterwards. 
I can see the failing integration test, only the timestamp for some reason, not the date...

yalwan-scee:
> 3: x = masked_array(data = [--],
> 3:              mask = [ True],
> 3:        fill_value = NaT)
> 3: 
> 3: y = masked_array(data = [datetime.datetime(400, 12, 31, 1, 2, 3)],
> 3:              mask = [False],
> 3:        fill_value = NaT)

It looks like numpy itself is ok, but the select pulled .. a null out, or something else?

yalwan-scee:
So, this was my concern:

> insert into fpqtest values(2, '0017-01-01 00:00:00', 'h', false);
> select * from fpqtest;
> 2	0017-01-01 00:00:00	h	False

Fine..

```
>>> D = cursor.fetchallnumpy()
>>> D['FIELD2'][0]
numpy.datetime64('0017-01-01T01:01:15.000000')

```

Date drifting is harmful for dates wayback, seems it is not as simple as naive division by 60*60*24? We somehow gained 3600 + 60 + 15 seconds .. but the date component is correct.

MathMagique:
Hm, that's unfortunate. Drifting dates seem more harmful than being limited by boost gregorian's date range. Suggestions anybody?

MathMagique:
I'll abort this because nobody has come up with a trivial way to fix this for a long time. Nevertheless, thanks a lot for the time you invested, @yaxxie

PR #69, opened 2017-03-27
Numpy batched fetch api

yaxxie:
This is an initial PR to get tests run and feedback against the implementation as it stands for batched numpy fetches, including a python call to do the same.

I rewrote `fetchallnumpy()` -- it looks the same to the user but under the hood is grabbing the generator and concatenating all the chunks together
I implemented `fetchbatchnumpy()` -- This immediately returns the next batch to be fetched
Both these methods are supported under the hood by the generator function `_numpy_batch_generator()` -- this returns a generator object which returns an array of masked arrays for each `next()`, without putting it into an ordered dict (this is left to the user facing methods to do)

I did not write any integration tests yet, as I would like some guidance on what you would like to see.

All feedback appreciated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ Version 1.1.0 (unreleased)
     In particular, a few `long` types were converted to `intptr_t` and `int64_t`
     where appropriate. In particular, this affects the `field` type that may be used
     by C++ end users (so they exist).
-*   Added new cursor.fetchnumpybatches method which returns a generator to
+*   Added new `cursor.fetchnumpybatches()` method which returns a generator to
     iterate over result sets in batch sizes as defined by buffer size or rowcount
 
 Version 1.0.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Version 1.1.0 (unreleased)
     In particular, a few `long` types were converted to `intptr_t` and `int64_t`
     where appropriate. In particular, this affects the `field` type that may be used
     by C++ end users (so they exist).
+*   Added new cursor.fetchnumpybatches method which returns a generator to
+    iterate over result sets in batch sizes as defined by buffer size or rowcount
 
 Version 1.0.5
 =============
@@ -100,7 +102,7 @@ Version 0.4.1
     (linking is already done by the Python interpreter). This was always
     the case for the libraries in the packages uploaded to PyPI, so no
     change was necessary here.
-*   Internal: Some modifications to the structure of the underlying 
+*   Internal: Some modifications to the structure of the underlying
     C++ code.
 
 Version 0.4.0
@@ -122,7 +124,7 @@ Version 0.3.0
     batch in the background. This may yield substantial performance improvements
     in the right circumstances (results are retrieved in roughly the same speed
     as they are converted to Python objects).
-    
+
     Ansynchronous I/O support is experimental. Enable it with
     `turbodbc.connect('My data source name', use_async_io=True)`
 
@@ -149,7 +151,7 @@ Version 0.2.3
 
 *   Fix issue that only lists were allowed for specifying parameters for queries
 *   Improve parameter memory consumption when the database reports very large
-    string parameter sizes 
+    string parameter sizes
 *   C++ backend: Provides more low-level ways to access the result set
 
 Version 0.2.2
--- a/README.md
+++ b/README.md
@@ -149,10 +149,23 @@ Here is how to execute an `INSERT` query with many parameters:
 NumPy support
 -------------
 
-Here is how to retrieve a result set in the form of NumPy arrays:
+Here is how to retrieve a full result set in the form of NumPy arrays:
 
     >>> cursor.execute("SELECT A, B FROM my_table")
     >>> cursor.fetchallnumpy()
+    OrderedDict([('A', masked_array(data = [42 --],
+                                    mask = [False True],
+                                    fill_value = 999999)),
+                 ('B', masked_array(data = [3.14 2.71],
+                                    mask = [False False],
+                                    fill_value = 1e+20))])
+
+You can also fetch numpy result sets in batches, based on the `read_buffer_size` attribute on the connection, using an iterable:
+
+    >>> cursor.execute("SELECT A, B FROM my_table")
+    >>> batches = cursor.fetchnumpybatches()
+    >>> for batch in batches:
+    ...     print(batch)
     OrderedDict([('A', masked_array(data = [42 --],
                                     mask = [False True],
                                     fill_value = 999999)),
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ Here is how to retrieve a full result set in the form of NumPy arrays:
                                     mask = [False False],
                                     fill_value = 1e+20))])
 
-You can also fetch numpy result sets in batches, based on the `read_buffer_size` attribute on the connection, using an iterable:
+You can also fetch NumPy result sets in batches, based on the `read_buffer_size` attribute on the connection, using an iterable:
 
     >>> cursor.execute("SELECT A, B FROM my_table")
     >>> batches = cursor.fetchnumpybatches()
--- a/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
+++ b/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
@@ -51,12 +51,6 @@ numpy_result_set::numpy_result_set(turbodbc::result_sets::result_set & base) :
 pybind11::object numpy_result_set::fetch_next_batch()
 {
 	std::size_t rows_in_batch = base_result_.fetch_next_batch();
-
-	if (!rows_in_batch)
-	{
-		throw pybind11::stop_iteration();
-	}
-
 	auto const column_info = base_result_.get_column_info();
 	auto const n_columns = column_info.size();
 
--- a/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
+++ b/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
@@ -48,10 +48,15 @@ numpy_result_set::numpy_result_set(turbodbc::result_sets::result_set & base) :
 {
 }
 
-
-pybind11::object numpy_result_set::fetch_all()
+pybind11::object numpy_result_set::fetch_next_batch()
 {
 	std::size_t rows_in_batch = base_result_.fetch_next_batch();
+
+	if (!rows_in_batch)
+	{
+		throw pybind11::stop_iteration();
+	}
+
 	auto const column_info = base_result_.get_column_info();
 	auto const n_columns = column_info.size();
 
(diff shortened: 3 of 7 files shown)


yaxxie:
Will fix the implementation so that the `test_numpy_empty_column` integration test does not break.

@MathMagique would you mind to let me know what else you would like to see here?

cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
```diff
@@ -48,10 +48,15 @@ numpy_result_set::numpy_result_set(turbodbc::result_sets::result_set & base) :
 {
 }
 
-
-pybind11::object numpy_result_set::fetch_all()
+pybind11::object numpy_result_set::fetch_next_batch()
 {
 	std::size_t rows_in_batch = base_result_.fetch_next_batch();
+
+	if (!rows_in_batch)
```
MathMagique: please use `not` instead of `!`, easier to read

python/turbodbc/cursor.py
```diff
@@ -122,14 +123,34 @@ def fetchmany(self, size=None):
         return [row for row in islice(self, size)]
 
     def fetchallnumpy(self):
+        if _has_numpy_support():
```
MathMagique: Seems like a duplicate since the same check is part of the batch generator

MathMagique:
The existing tests assert that the `fetchallnumpy()` functionality works as expected (and as before), so these tests should not need modification at all.

I would prefer the `fetchbatchnumpy()` functionality to return a generator. This would allow users to iterate over all batches in a natural way. The name should be adjusted to `fetchnumpybatches()` or similar, then. Still, it probably makes sense to keep the low-level generator object that does not perform the `OrderedDict` transformation.

I am missing some integration tests for the new batch functionality. I would like to see a test where you fetch a total of 5 rows (say integers `[1, 2, 3, 4, 5]`) in batches of two rows. Then check that the first batch contains `[1, 2]`, the next `[3, 4]`, and the last batch `[5]`. Also check there is no fourth batch. Checking a single data type should be sufficient, since all data types and null values are already checked by the previous `fetchallnumpy()` functionality (and thus tests the low-level implementation). Also, I'd like to see some tests with the new functionality for empty result sets.

For the tests, you can set the `rows_to_buffer` parameter to `2` when calling the `open_connection` or `open_cursor` context managers (https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc_test/helpers.py#L86). This should give you the desired level of control.

Also, I am missing some elementary documentation in the README file. An entry in the changelog would be great as well (probably version 1.1.1 will be the next one). Don't worry about the Appveyor build just now; this is work in progress.

cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
yaxxie: Sorry, I didn't realise this was a thing in c++11

MathMagique: It is not a C++11 thing, it's probably just my thing. I also don't like `&&` (use `and`) or `||` (use `or`). I'm just glad that Python got this part right ;-).

yaxxie:
Thanks for the feedback, will work on these and get back to you

cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
keitherskine: @MathMagique I like your thinking!  Clear code is key.  The fewer hieroglyphics the better.

yaxxie:
I got a little problem

In numpy_result_set.cpp we have this:
```
	if (not rows_in_batch)
	{
		throw pybind11::stop_iteration();
	}
```

This is problematic because there is no empty "typed" list being returned any longer, via 

```
	std::vector<std::unique_ptr<masked_column>> columns;
	for (std::size_t i = 0; i != n_columns; ++i) {
		columns.push_back(make_masked_column(column_info[i].type));
	}
```

so when we retrieve the result set, we don't get anything except a StopIteration exception. The reason why I have a problem with this, is that this differs in the previous behavior of returning empty **typed** arrays, which we cannot do any longer.

Should we
1) Not raise the stop iteration, initialise the empty list, and catch the fact that its empty in the python and modify the generator?
2) Not be concerned about this?
3) Do something else?

cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
yaxxie: I don't disagree, this opened my eyes (I had no idea this was legal in C++)

MathMagique:
Interesting. I'd opt for solution 1 you proposed. In the context of what was before, this makes sense. Also, I think the Python result set returns an empty list when no more results could be retrieved, so this would be consistent, at least.

[If no previous behavior had been given, your initial solution would have been acceptable behavior as well.]

yaxxie:
So, c71bbb6 was to do option 1, and the unit tests pass. I feel like the code is a bit messy as a consequence but if you have no complaints I'll leave it. Working on the additional integration tests and documentation, but I'm not sure where the changelog is.

MathMagique:
The changelog can be found in the root directory: `CHANGELOG.md`

yaxxie:
Please review 84b6aab, and the changes as a whole. Provided the new integration tests pass, are you happy with this as it stands or would you like to see anything else?

MathMagique:
@yaxxie I'll make the thorough review as soon as I can; probably not today.

yaxxie:
@MathMagique were you able to take a look yet?

MathMagique:
not yet, sorry :-(

python/turbodbc/cursor.py
```diff
@@ -7,23 +7,24 @@
...
 
-
 def _has_numpy_support():
+
     try:
         import turbodbc_numpy_support
         return True
     except ImportError:
-        return False
+         return False
```
MathMagique: The indentation here is one space too much :-D

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Version 1.1.0 (unreleased)
     In particular, a few `long` types were converted to `intptr_t` and `int64_t`
     where appropriate. In particular, this affects the `field` type that may be used
     by C++ end users (so they exist).
+*   Added new cursor.fetchnumpybatches method which returns a generator to
```
MathMagique: please use backticks and brackets to mark this as a function: `cursor.fetchnumpybatches()`

README.md
```diff
@@ -160,6 +160,19 @@ Here is how to retrieve a result set in the form of NumPy arrays:
                                     mask = [False False],
                                     fill_value = 1e+20))])
 
+You can also fetch numpy result sets in batches, based on the `read_buffer_size` attribute on the connection, using an iterable:
```
MathMagique: Please use the established way of writing NumPy in this document

python/turbodbc/cursor.py
```diff
@@ -7,23 +7,24 @@
...
-def _make_masked_array(data, mask):
+def _make_masked_arrays(result_batch):
     from numpy.ma import MaskedArray
     from numpy import object_
-    if isinstance(data, list):
-        return MaskedArray(data=data, mask=mask, dtype=object_)
-    else:
-        return MaskedArray(data=data, mask=mask)
-
+    numpy_masked_array = []
```
MathMagique: please use the plural form `numpy_masked_arrays` (or even leave the `numpy` part out)

python/turbodbc/cursor.py
```diff
@@ -122,17 +123,34 @@ def fetchmany(self, size=None):
...
-            return OrderedDict(columns)
         else:
             raise Error("turbodbc was compiled without numpy support. Please install "
                         "numpy and reinstall turbodbc")
+        numpy_result_set = make_numpy_result_set(self.impl.get_result_set())
+        firstrun = True
+        while True:
+            result_batch = numpy_result_set.fetch_next_batch()
+            batch_to_yield = _make_masked_arrays(result_batch)
+            next_batch = len(list(filter(len, batch_to_yield)))
```
MathMagique: I find the logic here a bit confusing. Basically, you want to know if this batch is empty or not. That could be done with a variable like this (unless I am missing something, obviously):
```
is_empty_batch = (len(batch[0]) == 0)
```

python/turbodbc/cursor.py
```diff
@@ -122,17 +123,34 @@ def fetchmany(self, size=None):
...
-                          [_make_masked_array(data, mask) for data, mask in numpy_result_set.fetch_all()])
-            return OrderedDict(columns)
         else:
             raise Error("turbodbc was compiled without numpy support. Please install "
                         "numpy and reinstall turbodbc")
+        numpy_result_set = make_numpy_result_set(self.impl.get_result_set())
+        firstrun = True
+        while True:
+            result_batch = numpy_result_set.fetch_next_batch()
+            batch_to_yield = _make_masked_arrays(result_batch)
```
MathMagique: `batch_to_yield` is a bit of an unwieldy name, probably because the good name was already taken a line above. You could combine both lines into one:
```
batch = _make_masked_arrays(numpy_result_set.fetch_next_batch())
```

python/turbodbc_test/test_select_numpy.py
```diff
@@ -48,9 +48,21 @@ def test_numpy_empty_column(dsn, configuration):
             cursor.execute("SELECT a FROM {}".format(table_name))
             results = cursor.fetchallnumpy()
             assert isinstance(results, OrderedDict)
-            assert len(results) == 1
+            assert len(results) == 1 # ncols
             assert isinstance(results[_fix_case(configuration, 'a')], MaskedArray)
 
+@for_each_database
+def test_numpy_empty_column_batch_fetch(dsn, configuration):
```
MathMagique: good test!

python/turbodbc_test/test_select_numpy.py
```diff
@@ -109,6 +121,19 @@ def test_numpy_binary_column_larger_than_batch_size(dsn, configuration):
...
 
+@for_each_database
+def test_numpy_batch_fetch(dsn, configuration):
+    with open_cursor(configuration, rows_to_buffer=2) as cursor:
+        with query_fixture(cursor, configuration, 'INSERT INTEGER') as table_name:
+            cursor.executemany("INSERT INTO {} VALUES (?)".format(table_name),
+                               [[1], [2], [3], [4], [5]])
+            cursor.execute("SELECT a FROM {} ORDER BY a".format(table_name))
+            batches = cursor.fetchnumpybatches()
+            expected_list = [MaskedArray([1, 2], mask=False), MaskedArray([3, 4], mask=False), MaskedArray([5], mask=False)]
```
MathMagique: please use line breaks to separate the batches visually:

```
expected_list = [MaskedArray([1, 2], mask=False),
                 MaskedArray([3, 4], mask=False),
                 MaskedArray([5], mask=False)]
```
Also, `expected_list` should probably be renamed to `expected_batches`

python/turbodbc/cursor.py
yaxxie: Very happy to change, but just to explain, hard coded indexes make me a little twitchy, so I sought a way to determine empty batched that did not use a hardcoded indexing, although I can't think of a time when a result set will not have even 1 column...

python/turbodbc_test/test_select_numpy.py
yaxxie: Thanks!

python/turbodbc/cursor.py
MathMagique: Yeah, indexes can be weird. Result sets are guaranteed to have at least one column, see this line of code: https://github.com/blue-yonder/turbodbc/blob/master/cpp/turbodbc/Library/src/command.cpp#L35

MathMagique:
Thanks @yaxxie for this wonderful contribution! It actually added flexibility, new usability patterns, _and_ it simplified the code. On second thought, this is probably how I should have done it the first time around ;-). Great job!

yaxxie:
Thanks for all the help, it was much appreciated!

PR #70, opened 2017-03-28
Try to improve CMake build on Windows

TWAC:
Uses CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS instead of GENERATE_EXPORT_HEADER.

touch and sh is usually not available on Windows, so I used cmake -E and bat instead.

Controlling the googletest build is quite awkward, have you considered something like [this](https://crascit.com/2015/07/25/cmake-gtest/)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ addons:
       - unixodbc
       - unixodbc-dev
       - libboost-date-time-dev
+      - libboost-locale-dev
       - libboost-system-dev
       - libmyodbc
       - odbc-postgresql=1:09.02.0100-2ubuntu1
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ build_script:
 
     cd turbodbc_build
 
-    cmake ..\turbodbc -DPYBIND11_PYTHON_VERSION=3.6 -A x64
+    cmake ..\turbodbc -DPYBIND11_PYTHON_VERSION=3.6 -DCMAKE_VERBOSE_MAKEFILE=ON -A x64
 
     cmake --build .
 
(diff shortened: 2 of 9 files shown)


MathMagique:
It's somewhat expected for the AppVeyor build to fail, but it is weird that the Travis build fails. I have retriggered an old build on the `windows_ci` branch to see whether something changed on travis or whether it is one of your changes that is the culprit.

MathMagique:
The problem seems to be the inclusion of boost locale in the `find_package` thing. The prebuilt library seems to be missing on Linux.

MathMagique:
Try adding `libboost-locale-dev` to https://github.com/blue-yonder/turbodbc/blob/master/.travis.yml#L56, this should fix the problem.

I have no idea why the build works on master without this requirement.

TWAC:
Thanks, libboost-locale-dev did the trick.

The travis builds are of type `Debug`, but it looks like the python debugging binaries are [missing](https://ci.appveyor.com/project/MathMagique/turbodbc/build/windows_ci%20-%20%2351#L908) on appveyor. So I switched that build to `Release`.

In my experience `/W4` just spams a lot of useless stuff, so I reduced it to `/W3`

Anyway, now it at least build, which will make it easier to check for regressions. Getting the cmake build to run and pass tests will be another hurdle.

MathMagique:
Yes, that's excellent! How to proceed? I am willing to do the merge and I would continue to get the tests running on Friday. Alternatively, if you would like to continue, I can leave this PR open for your convenience. What shall it be?

TWAC:
Please merge now, having other pull requests compile tested against Windows will be helpful.

I can open another PR if I get something working.

MathMagique:
@TWAC Thanks for your help!

PR #73, opened 2017-04-02
Windows ci

MathMagique:
This pull request enables official Windows support with a continuous integration build made with Appveyor.


MathMagique:
It seems I have managed to create a binary distribution of turbodbc for Windows: https://ci.appveyor.com/api/buildjobs/1hbdwaen0l8p5hkb/artifacts/turbodbc-1.1.1-cp36-cp36m-win_amd64.whl

It seems to work on my gaming PC, on which I have a copy of boost 1.63 installed. It also requires 64 bit Python and whatever runtime comes with Visual Studio Community Edition 2015 Update 3.

It would be great if someone could give this wheel a spin @keitherskine @TWAC :-D

TWAC:
The wheel works fine for me on Python 3.6.1 x64.

I would expect it to require vcredist 2015 installed, but not Boost.

keitherskine:
@MathMagique  I got your wheel to install by running `pip install` in an Administrator command prompt.  In a normal command prompt I got `PermissionError: [WinError 5] Access is denied: 'C:\\Program Files\\Python36\\Lib\\site-packages\\pybind11'`.

When trying to use `turbodbc`, though, I get the following (almost certainly because I don't have boost or Visual Studio installed):
```C:\WINDOWS\system32>"C:\Program Files\Python36\python.exe"
Python 3.6.1 (v3.6.1:69c0db5, Mar 21 2017, 18:41:36) [MSC v.1900 64 bit (AMD64)] on win32
>>> import turbodbc
  File "C:\Program Files\Python36\lib\site-packages\turbodbc\__init__.py", line 4, in <module>
    from .connect import connect
  File "C:\Program Files\Python36\lib\site-packages\turbodbc\connect.py", line 6, in <module>
    from turbodbc_intern import connect as intern_connect
ImportError: DLL load failed: The specified module could not be found.
```

So, err, what do I have to do to get past this error?  Do I have to install boost, or Visual Studio, or both?  If I have to install Visual Studio, can you give me an indication of what "workload" is required.  Apparently, you have to choose a [workload or individual components](https://docs.microsoft.com/en-us/visualstudio/install/install-visual-studio) as part of the install, and it would be kinda nice to install only what's absolutely necessary.

MathMagique:
@keitherskine Try downloading and installing the [Microsoft Visual Studio 2015 Update 3 Redistributable 64 bit](https://www.microsoft.com/en-us/download/details.aspx?id=53840).
Then please check again if this is sufficient to get turbodbc to run.
If this is not the case, the next thing would be to install [boost 1.63 64 bit compiled with Visual Studio 2015](https://sourceforge.net/projects/boost/files/boost-binaries/1.63.0/boost_1_63_0-msvc-14.0-64.exe/download). Then check again, please.

The Visual Studio redistributable clocks in at just 14 MB, Boost requires about 300 MB to download.

keitherskine:
@MathMagique Installing _MSVS 2015 update 3_ did the trick (even though it doesn't appear to be certified for Windows 10).  I didn't even need to re-install turbodbc.  No need to install boost either.  Many thanks!

MathMagique:
@keitherskine Thank _you_ for letting me know what to put in the Windows installation instructions section ;-).

PR #74, opened 2017-04-03
Start testing MSSQL and Python 3.5 on appveyor

TWAC:
Checking appveyor


appveyor.yml
```diff
@@ -49,14 +58,20 @@ test_script:
 
     reg import ..\turbodbc\appveyor\odbc-postgresql.reg
 
+    reg import ..\turbodbc\appveyor\odbc-mssql.reg
+
     reg import ..\turbodbc\appveyor\odbc-data-sources.reg
 
     SET PATH=C:\Program Files\PostgreSQL\9.6\bin;C:\Program Files\MySQL\MySQL Server 5.7\bin;%PATH%
 
+    python -c "open('../turbodbc/python/turbodbc_test/mssql.json','w').write(open('../turbodbc/python/turbodbc_test/query_fixtures_mssql.json').read().replace('StrongPassword1','Password12!'))"
```
MathMagique: I really don't like this line :-). Either change the MSSQL password to be consistent with the existing test fixture, or change the password listed in the test fixture to be consistent with the MSSQL password.

TWAC: OK, I will change it to something like `sqlcmd -S localhost\SQL2016 -U SA -P "Password12!" -Q "ALTER LOGIN SA WITH PASSWORD = 'StrongPassword1'"`. But first I will let the current build jobs run to completion.

MathMagique: Sure :-).

TWAC:
More testing to catch regressions, also produces binaries for Python 3.5 on Windows.

The 3.6 and 3.5 builds run serially, so this unfortunately doubles the testing time.

MathMagique:
Nice work!

PR #75, opened 2017-04-04
add option to use preformatted ODBC connection string

TWAC:
Moving from ceODBC to turbodbc can be quite annoying since ceODBC only supports preformatted strings.
It also looks like pyodbc [currently recommends](https://github.com/mkleehammer/pyodbc/wiki/Connecting-to-databases) preformatted strings.
This makes it easier.


MathMagique:
Code looks good with the following exceptions:
* Documentation in README file is missing
* I'd like to see a test that actually uses the connection string to establish a connection. The connection string could be constructed directly from the information contained within the fixture files (dsn and, optionally, user and password).

TWAC:
Yes, that sounds reasonable.

MathMagique:
Thanks for adding this!

PR #79, opened 2017-04-10
docs: contributing, remove obsolete step 6

keitherskine:
Step 6 in the description of how to create a development version is now obsolete. I also changed the numbering from hard-coded to automatic.

(I thought it would be cleaner to create a new PR).


MathMagique:
History is supposed to be messy, I hear ;-)

Thanks a lot!

Issue #77, opened 2017-04-10
Feature-Request: Release GIL


chmp:
As a (generic) user, I want turbodbc to release the GIL to allow concurrent usage.

Currently, turbodbc does not release the GIL. Thereby, it is not possible to use python threading effectively with turbodbc. By releasing the GIL, turbodbc would allow running multiple queries concurrently or execute queries in the background of an jupyter notebook without locking it completely. For example, [psycopg](http://initd.org/psycopg/docs/) does release the GIL.

chmp:
I'll give it a try and report back how it goes :). Small caveat: I'm mostly interested in the `fetchallnumpy` code path and won't look much at other place (performance and all).

MathMagique:
Fixed in #80

yaxxie:
Just FYI, just came across this, and am actually happy because i'm about to test using this in threaded fashion in my project. If this works as I expect I can double throughput of my application, so I'd like to add my thanks for this contribution.

PR #80, opened 2017-04-10
WIP: Release GIL

chmp:
My initial implementation of #77. I focused on the functions I assumed most time will be spent in without access to python: `execute` and `fetch_next_batch`. Something, I was wondering while playing around with my version: turbodbc itself is not threadsafe, is it?

Currently, I'm testing how it holds up in practice. Initial tests are promising (after implementing a shoddy connection pool).

### DIFF
--- a/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
+++ b/cpp/turbodbc_numpy/Library/src/numpy_result_set.cpp
@@ -14,62 +14,62 @@ namespace turbodbc_numpy {
 
 namespace {
 
-	std::unique_ptr<masked_column> make_masked_column(turbodbc::type_code type)
-	{
-		switch (type) {
-			case turbodbc::type_code::floating_point:
-				return std::unique_ptr<binary_column>(new binary_column(numpy_double_type));
-			case turbodbc::type_code::integer:
-				return std::unique_ptr<binary_column>(new binary_column(numpy_int_type));
-			case turbodbc::type_code::boolean:
-				return std::unique_ptr<binary_column>(new binary_column(numpy_bool_type));
-			case turbodbc::type_code::timestamp:
-			case turbodbc::type_code::date:
-				return std::unique_ptr<datetime_column>(new datetime_column(type));
-			case turbodbc::type_code::unicode:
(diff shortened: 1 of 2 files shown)


chmp:
After looking at the diff, I just saw some tabs / spaces problems. I will fix those.

MathMagique:
When in doubt, just replace tabs with four spaces in the files you modified. It's what I currently do when I modify files.

If I am right, this will not modify performance at all, just what you can do in another thread at the same time? If so, an integration test for the new behavior would be nice to prevent accidental deletion ;-).

MathMagique:
To your question: Turbodbc is only thread-safe when it comes to connecting and disconnecting. Connections are not intended to be shared among several threads. See the definition of the `threadsafety` constant [here](https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc/api_constants.py#L6) and [what it means](https://www.python.org/dev/peps/pep-0249/#threadsafety). Sharing connections might still be possible, but since I do not have any tests for that I cannot give any guarantees (which explains the current value of the `threadsafety` constant).

chmp:
Thanks for the explanation of the thread safety guarantees. It's well appreciated, since I'm not quite aware of the intricacies of the python DB API. Your explanation confirms my observation when I reused the connection across threads: I executed two queries concurrently, both were shown as running, but postgres only reported a single query active. 

I retabbed the files, as per your suggestion. Somewhat destroys diffs, but maybe the individual commits are of use.

With regard to the integration test, I may have an idea: in postgres has a function called `pg_sleep`. Trying to execute python code concurrently, while the db is waiting, should allow to test for the GIL. However, this setup seems somewhat fickle, since it really depends on execution times and order. Not sure it's worth the hassle. What are your thoughts?

chmp:
Thanks for merging.

MathMagique:
@chmp Hi Christopher! Sorry to have kept you waiting so long. I arrived at the conclusion that it would not be fair to have you implement the first nonfunctional test for turbodbc. It would not suffice to make the test pass with a single database, but the test should be executable with all databases. I guess the value of providing the test does not really justify the required effort at the moment. So I merged your request. Thanks for contributing!

PR #83, opened 2017-04-15
api reference doc improvements

keitherskine:
- formatting, the parameter section sometimes was not separated out
- remove unnecessary line continuation characters
- wording

I've built the relevant HTML docs using sphinx-build and they look OK to me.

### DIFF
--- a/python/turbodbc/cursor.py
+++ b/python/turbodbc/cursor.py
@@ -60,12 +60,12 @@ class Cursor(object):
     @property
     def description(self):
         """
-        Retrieve a description of the columns in the current result set
-        
-        :return: A tuple of seven elements. Only some elements are meaningful:\n
+        Retrieve a description of the columns in the current result set.
+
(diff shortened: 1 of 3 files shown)


python/turbodbc/cursor.py
```diff
@@ -60,12 +60,12 @@ def _assert_valid_result_set(self):
...
-        Retrieve a description of the columns in the current result set
-        
-        :return: A tuple of seven elements. Only some elements are meaningful:\n
+        Retrieve a description of the columns in the current result set.
+
+        :return: A tuple of seven elements (or just ``None`` if no result set exists). Only some elements are meaningful:\n
                  *   Element #0 is the name of the column
                  *   Element #1 is the type code of the column
-                 *   Element #6 is true if the column may contain ``NULL`` values
+                 *   Element #6 is true if the column may contain `NULL` values
```
MathMagique: I think the two backticks were actually correct (rst formatting)

keitherskine:
I might have got a little carried away with this PR so best if I take another stab at it.

Issue #89, opened 2017-05-18
Distribution of more flavours of wheels


FlorianWilhelm:
It would be really nice to have also wheels for Linux (`manylinux1_x86_64`) and (`cp34m`) for Turbodbc on PyPI. Many Cloudera and in general Linux distributions only provide Python 3.4. Thanks a bunch for your great work!

MathMagique:
Hi Florian! You are absolutely right, those wheels would make installing turbodbc that much easier.

xhochy:
@MathMagique We would then build wheels that have `unixodbc` bundled? Then I could build a quick Dockerfile that will emit these wheels and also make macOS wheels via a separate script.

@FlorianWilhelm For the time being, I've updated the `conda` packages to the latest version: https://github.com/conda-forge/turbodbc-feedstock / https://anaconda.org/conda-forge/turbodbc

MathMagique:
@xhochy Yes, I think that would be the most reliable thing to do. Many Linux variants use unixodbc 2.3.x in the meantime (Ubuntu 16+, Debian 8+), so I would suggest to bundle the latest officially released version (2.3.4) with the wheels.

Probably it makes sense to add some build information to turbodbc so users can query the unixodbc version used during the build.

MathMagique:
Hi nrhy! Sadly, the conversation did not continue with much momentum. In the meantime, you can use the `conda` package manager to install the precompiled [turbodbc conda packages](https://anaconda.org/conda-forge/turbodbc).

nrhy:
Unfortunately conda is not supported in my workspace :(

MathMagique:
That's unfortunate. In this case, I guess that until official manylinux wheels can be provided, you would need to create wheels for your specific operating system (`pip wheel` should help). Just make sure you build the wheel in an environment with NumPy/Arrow installed so these add-ons are built as well.

You can then distribute the wheel among your colleagues, e.g., via [devpi](https://devpi.net/docs/devpi/devpi/stable/%2Bd/index.html) or file transfer. If they share the same OS, you should be safe, although not very convenient.

nrhy:
@MathMagique Yeah, it's what I've been doing. It'll have to do for now.

Thank you!

mpeleshenko:
+1 on this. Having official Windows + Linux wheels available on pypi.org for at least Python 3.5+ would make installing turbodbc much easier. I remember when pyarrow starting putting out wheels on pypi in addition to their conda-forge builds.

[...]

PR #108, opened 2017-06-28
WIP: Build manylinux1 wheels

xhochy:
Depends on https://github.com/pypa/auditwheel/issues/76

### DIFF
--- a/contrib/manylinux1/.gitignore
+++ b/contrib/manylinux1/.gitignore
@@ -1,0 +1,1 @@
+turbodbc
--- a/contrib/manylinux1/Dockerfile-x86_64
+++ b/contrib/manylinux1/Dockerfile-x86_64
@@ -1,0 +1,20 @@
+FROM quay.io/pypa/manylinux1_x86_64:latest
+
+ADD scripts/build_boost.sh /
+RUN /build_boost.sh
+
+ADD scripts/build_unixodbc.sh /
+RUN /build_unixodbc.sh
+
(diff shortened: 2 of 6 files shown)


MathMagique:
Relates to #89

yaxxie:
Hi guys,

I was working on manually grafting a manylinux wheel for distribution within my org. I noticed this as well (and decided it as acceptable), but I am just wondering what it is about pyarrow that auditwheel wants to graft `libarrow_python.so`, and yet it doesn't want to graft the numpy libraries. It makes me wonder if it is more an issue with the way pyarrow is built and perhaps there is something that should be changed upstream there instead?

On the other hand, I struggled with the `RPATH` for the other shared objects that point to `libturbodbc.so` because once their `RPATH` is updated for the vendored shared objects, `libturbodbc.so` is no longer in the `RPATH` and is not found at runtime. I am working around this by hardlinking `libturbodbc.so` into the vendor paths at the same time as hardlinking the vendored shared objects for `libturbodbc.so` (because libodbc.so.2 and libltdl.so.3 are not being vendored for this one), as well as fixing a libname change which doesn't look like it worked correctly against `turbodbc_intern.so`.

I got a working (import turbodbc works, didn't test properly) manylinux wheel like this, which is obviously fugly, but I'm wondering if you have any tips or comments to add?

yaxxie:
I guess I should try your dockerfile that I never noticed ...

PR #125, opened 2017-09-08
Add support for Apache Arrow parameters 

xhochy:
This depends on the yet unreleased `pyarrow==0.7.0`. Would be nice to get a review before the actual release.


xhochy:
@MathMagique Code formatting is not yet in a good shape in this PR. Instead of manually fixing this, I would rather add `clang-format` to the build toolchain to automate this.

xhochy:
@MathMagique I'm now at the point where the CMake pre-link step fails with 

```
"setlocal
cd C:\projects\turbodbc_build
if %errorlevel% neq 0 goto :cmEnd
"C:\Program Files (x86)\CMake\bin\cmake.exe" -E __create_def C:/projects/turbodbc_build/cpp/turbodbc_arrow/Library/turbodbc_arrow_support.dir/Release/exports.def C:/projects/turbodbc_build/cpp/turbodbc_arrow/Library/turbodbc_arrow_support.dir/Release//objects.txt 
if %errorlevel% neq 0 goto :cmEnd
:cmEnd
endlocal & call :cmErrorLevel %errorlevel% & goto :cmDone
:cmErrorLevel
exit /b %1
:cmDone
if %errorlevel% neq 0 goto :VCEnd
:VCEnd" exited with code -1073741819.
```

This could either be a permission issue, a path length issue, or something else.

xhochy:
@MathMagique Ready for review!

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+            parameter_index(parameter_index)
+        {}
+
+        cpp_odbc::multi_value_buffer & get_buffer() {
+            return parameters.get_parameters()[parameter_index]->get_buffer();
+        }
+
+        template <size_t element_size>
+        void set_indicator(cpp_odbc::multi_value_buffer & buffer, int64_t start, int64_t elements) {
+            if (data->num_chunks() != 1) {
```
MathMagique: This check is made many times in the code. Please consider making the check elsewhere (in calling code, for example) or put the check in the parameter_converter class once and use the NVI pattern to enforce the check for subclasses

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+      template <typename String>
+        void set_batch_of_type(std::size_t start, std::size_t elements)
+        {
+          if (data->num_chunks() != 1) {
+            throw turbodbc::interface_error("Chunked string columns are not yet supported");
+          }
+
+          auto const& typed_array = static_cast<const BinaryArray&>(*data->chunk(0));
+
+          int32_t maximum_length = 0;
```
MathMagique: please extract this block to a function `determine_maximum_length()`, `rebind_buffer_to_match_maximum_length()`, or similar

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+          for (int64_t i = 0; i != elements; ++i) {
+            auto element = buffer[i];
+            if (typed_array.IsNull(start + i)) {
+              element.indicator = SQL_NULL_DATA;
+            } else {
+              int32_t out_length;
+              const uint8_t *value = typed_array.GetValue(start + i, &out_length);
+              std::memcpy(element.data_pointer, value, out_length);
+              element.indicator = character_size * typed_array.value_length(start + i);
```
MathMagique: I guess you should use out_length instead?

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+          auto & buffer = get_buffer();
+          auto const character_size = sizeof(typename String::value_type);
+
+          for (int64_t i = 0; i != elements; ++i) {
+            auto element = buffer[i];
+            if (typed_array.IsNull(start + i)) {
+              element.indicator = SQL_NULL_DATA;
+            } else {
+              int32_t out_length;
+              const uint8_t *value = typed_array.GetValue(start + i, &out_length);
```
MathMagique: please write this as `uint8_t const *`

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+            }
+        }
+
+        return converters;
+    }
+}
+
+void set_arrow_parameters(turbodbc::bound_parameter_set & parameters, pybind11::object const & pyarrow_table) {
+  arrow::py::import_pyarrow();
+  if (arrow::py::is_table(pyarrow_table.ptr())) {
```
MathMagique: No need for this check since in `cursor.py` you already check that you pass a table

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+    std::shared_ptr<Table> table;
```
MathMagique: please extract this into a function that returns `std::shared_ptr<Table>`

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -0,0 +1,375 @@
...
+    if (not arrow::py::unwrap_table(pyarrow_table.ptr(), &table).ok()) {
+      throw turbodbc::interface_error("Could not unwrap the C++ object from Python pyarrow.Table");
+    }
+
+    if (static_cast<int32_t>(parameters.number_of_parameters()) != table->num_columns()) {
```
MathMagique: please extract this into a function like `assert_table_columns_match_parameters()` or similar

python/turbodbc_test/test_executemanycolumns.py
```diff
@@ -131,71 +169,102 @@ def test_datetime64_nanoseconds_column(dsn, configuration):
...
+            if column_backend == 'arrow':
+                columns = [pa.Array.from_pandas(x) for x in columns]
+                columns = pa.Table.from_arrays(columns, ['column1', 'column2'])
             cursor.executemanycolumns("INSERT INTO {} VALUES (?, ?)".format(table_name), columns)
 
             results = cursor.execute("SELECT A, B FROM {} ORDER BY A".format(table_name)).fetchall()
             assert results == [[17, 3], [23, 2], [42, 1]]
 
 
+@for_each_column_backend
```
MathMagique: I have moved corner cases such as this to a separate test file: https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc_test/test_executemanycolumns_corner_cases.py#L64

setup.py
```diff
@@ -209,7 +209,7 @@ def get_extension_modules():
       packages = ['turbodbc'],
       extras_require={
           # We pin Apache Arrow quite restrictively until they guarantee a stable API
-          'arrow': ['pyarrow>=0.6,<0.7'],
+          'arrow': ['pyarrow>=0.7,<0.8'],
```
MathMagique: Any information when the arrow API will be reasonably stable?

xhochy:
@MathMagique If you really want tests for the rejection of chunked columns, I can do this but I don't see great value in that and it would delay the merge of the PR quite a lot.

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
MathMagique: The highest point would probably be the Python layer. That is also the place where checks for the related `c_contiguous` flag is done for NumPy arrays: https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc/cursor.py#L55

The Python layer also has the advantage of being a single code path (instead of six) that is easily testable with a single test ;-).

MathMagique:
Many thanks for your contribution!

PR #126, opened 2017-09-08
Support Dictionary-encoded String columns as Arrow output

xhochy:


cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -125,14 +130,26 @@ Status append_to_date_builder(size_t rows_in_batch, std::unique_ptr<ArrayBuilder
...
-    auto typed_builder = static_cast<StringBuilder*>(builder.get());
-    for (std::size_t j = 0; j != rows_in_batch; ++j) {
-        auto const element = input_buffer[j];
-        if (element.indicator == SQL_NULL_DATA) {
-            ARROW_RETURN_NOT_OK(typed_builder->AppendNull());
-        } else {
-            ARROW_RETURN_NOT_OK(typed_builder->Append(element.data_pointer, element.indicator));
+Status append_to_string_builder(size_t rows_in_batch, std::unique_ptr<ArrayBuilder> const& builder, cpp_odbc::multi_value_buffer const& input_buffer, uint8_t*, bool strings_as_dictionary) {
+    if (strings_as_dictionary) {
+        auto typed_builder = static_cast<StringDictionaryBuilder*>(builder.get());
```
MathMagique: How about extracting the blocks of the `if` blocks to a template function in the anonymous namespace? It is the same code, the only difference is the type of the `typed_builder`.

cpp/turbodbc_arrow/Library/src/python_bindings.cpp
```diff
@@ -20,5 +21,6 @@ PYBIND11_MODULE(turbodbc_arrow_support, module)
     pybind11::class_<arrow_result_set>(module, "ArrowResultSet")
         .def("fetch_all", &arrow_result_set::fetch_all);
 
-    module.def("make_arrow_result_set", make_arrow_result_set);
+    module.def("make_arrow_result_set", make_arrow_result_set,
+        pybind11::arg("result_set"), pybind11::arg("strings_as_dictionary") = false);
```
MathMagique: The default is set at the Python level in `cursor.py`. There is no need to specify the default a second time here, as it burdens to code with additional pybind11 options and introduces the risk of diverging defaults.

cpp/turbodbc_arrow/Library/turbodbc_arrow/arrow_result_set.h
```diff
@@ -46,6 +46,7 @@ class PYBIND11_EXPORT arrow_result_set {
 
 private:
 	turbodbc::result_sets::result_set & base_result_;
+  bool strings_as_dictionary_;
```
MathMagique: indentation seems off here, sorry for not having a formatting directive yet.

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -114,7 +116,7 @@ TEST_F(ArrowResultSetTest, SimpleSchemaConversion)
         "int_column", turbodbc::type_code::integer, size_unimportant, true}};
     EXPECT_CALL(rs, do_get_column_info()).WillRepeatedly(testing::Return(expected));
 
-    turbodbc_arrow::arrow_result_set ars(rs);
+    turbodbc_arrow::arrow_result_set ars(rs, false);
```
MathMagique: Instead of using `true` and `false` in the tests, please consider creating and using constants like
```
bool const strings_as_strings = false;
bool const strings_as_dictionaries = true;
```
That should make the tests more readable.

python/turbodbc/cursor.py
```diff
@@ -289,11 +289,11 @@ def _numpy_batch_generator(self):
             first_run = False
             yield result_batch
 
-    def fetchallarrow(self):
+    def fetchallarrow(self, strings_as_dictionary=False):
```
MathMagique: Could you please add a docstring for this function, so that it appears in the reference?

MathMagique:
Please also add a changelog entry and extend the documentation (advanced usage) to reflect the new option

MathMagique:
Excellent! Thanks @xhochy

PR #131, opened 2017-09-28
Fetch storage-optimal integer types in Arrow

### LINKED ISSUE #129: Add option to fetch optimal integer type for a column using Arrow
Instead of always fetching a column as INT64, we could also return the smallest possible integer type that suits for a column. This might not be the fastest, but the most memory efficient one.

xhochy:
Fixes #129

### DIFF
--- a/cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
+++ b/cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
@@ -27,6 +27,17 @@ namespace {
         MOCK_CONST_METHOD0(do_get_column_info, std::vector<turbodbc::column_info>());
         MOCK_CONST_METHOD0(do_get_buffers, std::vector<std::reference_wrapper<cpp_odbc::multi_value_buffer const>>());
     };
+
+    template <typename ArrowType>
+    void make_int_range(int64_t size, std::shared_ptr<arrow::Array>* out) {
+        typename arrow::TypeTraits<ArrowType>::BuilderType builder;
+        for (int64_t i = 0; i < size; i++) {
+          ASSERT_OK(builder.Append(i));
+        }
+        std::shared_ptr<arrow::Array> array;
+        ASSERT_OK(builder.Finish(out));
+    }
+
 }
 
 class ArrowResultSetTest : public ::testing::Test {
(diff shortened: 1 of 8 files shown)


cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -212,7 +215,29 @@ TEST_F(ArrowResultSetTest, MultiBatchConversionInteger)
             {"nonnull_int_column", turbodbc::type_code::integer, size_unimportant, false}});
     MockOutput({{BufferFromPrimitive(array, OUTPUT_SIZE, 0), BufferFromPrimitive(nonnull_array, OUTPUT_SIZE, 0)},
             {BufferFromPrimitive(array, OUTPUT_SIZE, OUTPUT_SIZE), BufferFromPrimitive(nonnull_array, OUTPUT_SIZE, OUTPUT_SIZE)}});
-    CheckRoundtrip(strings_as_strings);
+    CheckRoundtrip(strings_as_strings, plain_integers);
+}
+
+TEST_F(ArrowResultSetTest, ConversionCompressedInteger)
```
MathMagique: I have a really hard time reading this test case. The test intent is unclear for a number of reasons:
* The main assertion (that is called "CHECKRoundtrip") relies on member variables of a fixture.
* The checked roundtrip is not really a roundtrip. A roundtrip would be to insert the data and retrieve it.
* The test looks like it is about testing `Int64Builder` and `Int16Builder` because of all the extra asserts
* The terms `array` and `compressed_array` are not saying much. One could be called `input_64_bits` and the other `expected_output_16_bits`

The test would be much easier to read as this:
```
TEST_F(ArrowResultSetTest, ConversionCompressedInteger)
{
    auto input_data = make_64bit_integers_table(1024);
    auto base_result_set = make_prepared_results(input_data, {{"int_column", turbodbc::type_code::integer, size_unimportant, true}}, input_data);
    
    auto expected = make_16bit_integers_table(1024);
    assert_fetched_result(expected, base_result_set, strings_as_strings, compressed_integers);
}
```
(with adequately designed helper functions).

This layout would not require the use of a test fixture, makes important assertions explicit, and hides setup complexity (and setup assertions) in helper functions so the end result is easier to read.

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -220,23 +231,18 @@ TEST_F(ArrowResultSetTest, MultiBatchConversionInteger)
...
-      ASSERT_OK(compressed_builder.Append(i));
-    }
     std::shared_ptr<arrow::Array> array;
-    ASSERT_OK(builder.Finish(&array));
-    std::shared_ptr<arrow::Array> compressed_array;
-    ASSERT_OK(compressed_builder.Finish(&compressed_array));
-
-    expected_arrays.push_back(compressed_array);
-    expected_fields.push_back(arrow::field("int_column", arrow::int16(), true));
+    make_int_range<arrow::Int64Type>(1024, &array);
```
MathMagique: Why don't you just return a new shared pointer that contains an array?

python/turbodbc_test/test_select_arrow.py
```diff
@@ -72,6 +72,20 @@ def test_arrow_int_column(dsn, configuration):
 
 @for_each_database
 @pyarrow
+def test_arrow_int_column_adaptive(dsn, configuration):
+    with open_cursor(configuration) as cursor:
+        cursor.execute("SELECT 42 AS a")
+        result = cursor.fetchallarrow(adaptive_integers=True)
+        assert isinstance(result, pa.Table)
+        assert result.num_columns == 1
```
MathMagique: I guess the only really valuable assertions are the `int8` type checks and the `42` result. Everything else is sufficiently tested elsewhere.

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
xhochy: Because GTEST assertions only work in functions that return void.

MathMagique:
Thanks @xhochy

PR #135, opened 2017-10-09
Interpret Arrow results as stolen references

xhochy:
This fixes the current memory leak on using Arrow results.

### DIFF
--- a/python/turbodbc_test/test_select_arrow.py
+++ b/python/turbodbc_test/test_select_arrow.py
@@ -2,7 +2,9 @@ from collections import OrderedDict
 from mock import patch
 
 import datetime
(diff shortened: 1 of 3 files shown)


python/turbodbc_test/test_select_arrow.py
```diff
@@ -54,6 +56,8 @@ def test_arrow_empty_column(dsn, configuration):
             assert isinstance(result, pa.Table)
             assert result.num_columns == 1
             assert result.num_rows == 0
+            gc.collect()
```
MathMagique: I'd prefer the refcount check in a separate test with a clear name so that this test has a single intent.

Issue #149, opened 2018-02-22
PEP343 conformance


AtomBaf:
Hi blueyonder

It would be nice to be able to have a [PEP343 ](https://www.python.org/dev/peps/pep-0343/) compliant turbodbc connection.

This way, I could write something like:
`
with turbodbc.connect('...') as conn:
`

Thank you for your useful and great project

MathMagique:
Hi there! Makes sense to me, and I have the feeling that this would make for a nice and easy contribution :-).

PR #151, opened 2018-02-23
added PEP-343 conformance (Closes blue-yonder/turbodbc#149)

AtomBaf:


MathMagique:
Hi! Thanks for the contribution! Could you please add unit tests for the functionality you added? I'd suggest to use a connection in a context manager successfully, and then assert that using the connection outside of the context manager raises an exception. You could use https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc_test/test_connection.py#L10 as a template for a new test.
Similarly, a test for the cursor context could be put in https://github.com/blue-yonder/turbodbc/blob/master/python/turbodbc_test/test_cursor_basics.py#L22.

AtomBaf:
Here are the tests. Hopefully it will be easy to merge.

python/turbodbc_test/test_connection.py
```diff
@@ -151,3 +151,13 @@ def test_autocommit_querying(dsn, configuration):
     connection.autocommit = True
     assert connection.autocommit == True
     connection.close()
+
+@for_one_database
+def test_pep343_with_statement(dsn, configuration):
+
+    with connect(dsn, **get_credentials(configuration)) connection:
```
MathMagique: This `with` statement is missing an `as`, that's why the build breaks

MathMagique:
I just made the necessary change.

MathMagique:
Thanks for your contribution!

Issue #64, opened 2017-03-23
Issue with decoding UTF8 strings from WCHAR buffers


yalwan-scee:
If a field in a table contains (some, but not just any) Japanese characters, selecting these fields results in the following stacktrace:

```
SystemError: /builddir/build/BUILD/Python-3.5.3/Objects/listobject.c:307: bad argument to internal function

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/exceptions.py", line 31, in wrapper
    return f(*args, **kwds)
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/cursor.py", line 113, in fetchall
    return [row for row in self]
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/cursor.py", line 113, in <listcomp>
    return [row for row in self]
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/cursor.py", line 39, in __next__
    element = self.fetchone()
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/exceptions.py", line 31, in wrapper
    return f(*args, **kwds)
  File "/home/yalwan/.virtualenvs/fastpq/lib/python3.5/site-packages/turbodbc/cursor.py", line 105, in fetchone
    result = self.result_set.fetch_row()
SystemError: <built-in method fetch_row of PyCapsule object at 0x7fc347b3d360> returned a result with an error set
```

Examples of Japanese strings known to cause the issue:

ソニー・コンピュータエンタテインメント
ソニー・ピクチャーズ エンタテインメント
ベセスダ・ソフトワークス／ゼニマックス・アジア(株)

MathMagique:
Hi Yaqub! Could you please state the database, ODBC driver, ODBC driver settings, turbodbc version, and table scheme for future reference?

yalwan-scee:
turbodbc == 1.0.5
Python == 3.5.2
Database == IBM Puredata for Analytics
Table schema -- it can be reproduced using an nvarchar field

---------------------- odbc settings ---------------------------------
[NZSQL]
Driver                     = /usr/local/nz/lib/libnzsqlodbc3.so
[...]

PR #154, opened 2018-03-15
Adding fix unicode allocation parameter  (related to #64)

yaxxie:
add fix_unicode_allocation parameter which changes the way memory gets allocated for strings if the ODBC driver presents the string length field rather then the number of bytes required

Sorry if some of the whitespace changes aren't to your liking, I'll fix them as required (it is my text editors defaults).

I want to note that I didn't do anything to do with unit tests so far (although I tested that the changes function as expected manually). 

I also appreciate that this implementation might not be as you are expecting, happy to receive feedback on that too. 

Other pull request for changes to decode UTF8 from WCHAR buffers will also come in due course.

### DIFF
--- a/cpp/turbodbc/Library/src/make_description.cpp
+++ b/cpp/turbodbc/Library/src/make_description.cpp
@@ -102,11 +102,11 @@ template <typename Description>
 std::unique_ptr<description> make_character_description(cpp_odbc::column_description const & source,
                                                         turbodbc::options const & options)
 {
-    int multiplier = (options.fix_unicode_allocation ? Description::multiplier : 1);
+    int const capacity_multiplier = (options.force_extra_capacity_for_unicode ? Description::max_code_units_per_unicode_code_point : 1);
     std::size_t const sanitized_size = (source.size == 0 ? options.varchar_max_character_limit : source.size);
     std::size_t const limited_size = (options.limit_varchar_results_to_max ?
                                       std::min(sanitized_size, options.varchar_max_character_limit) :
-                                      sanitized_size) * multiplier;
+                                      sanitized_size) * capacity_multiplier;
     return std::unique_ptr<description>(new Description(source.name,
                                                         source.allows_null_values,
                                                         limited_size));
--- a/cpp/turbodbc/Library/src/make_description.cpp
+++ b/cpp/turbodbc/Library/src/make_description.cpp
@@ -9,6 +9,7 @@
 #include <sstream>
 #include <algorithm>
 
+
 namespace turbodbc {
 
 namespace {
@@ -101,10 +102,11 @@ template <typename Description>
 std::unique_ptr<description> make_character_description(cpp_odbc::column_description const & source,
                                                         turbodbc::options const & options)
 {
+    int multiplier = (options.fix_unicode_allocation ? Description::multiplier : 1);
     std::size_t const sanitized_size = (source.size == 0 ? options.varchar_max_character_limit : source.size);
     std::size_t const limited_size = (options.limit_varchar_results_to_max ?
                                       std::min(sanitized_size, options.varchar_max_character_limit) :
-                                      sanitized_size);
+                                      sanitized_size) * multiplier;
     return std::unique_ptr<description>(new Description(source.name,
                                                         source.allows_null_values,
                                                         limited_size));
--- a/cpp/turbodbc/Library/turbodbc/descriptions/string_description.h
+++ b/cpp/turbodbc/Library/turbodbc/descriptions/string_description.h
@@ -9,7 +9,7 @@ namespace turbodbc {
  */
 class string_description : public description {
 public:
-	const static int multiplier = 4;
+	const static int max_code_units_per_unicode_code_point = 4;
 	string_description(std::size_t maximum_length);
 	string_description(std::string name, bool supports_null, std::size_t maximum_length);
 	~string_description();
--- a/cpp/turbodbc/Library/turbodbc/descriptions/string_description.h
+++ b/cpp/turbodbc/Library/turbodbc/descriptions/string_description.h
@@ -9,6 +9,7 @@ namespace turbodbc {
  */
 class string_description : public description {
 public:
+	const static int multiplier = 4;
 	string_description(std::size_t maximum_length);
 	string_description(std::string name, bool supports_null, std::size_t maximum_length);
 	~string_description();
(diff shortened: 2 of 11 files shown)


yaxxie:
@MathMagique I think I did it right :) Let me know if there's something missing

cpp/turbodbc/Library/src/make_description.cpp
```diff
@@ -101,10 +102,11 @@ template <typename Description>
 std::unique_ptr<description> make_character_description(cpp_odbc::column_description const & source,
                                                         turbodbc::options const & options)
 {
+    int multiplier = (options.fix_unicode_allocation ? Description::multiplier : 1);
     std::size_t const sanitized_size = (source.size == 0 ? options.varchar_max_character_limit : source.size);
     std::size_t const limited_size = (options.limit_varchar_results_to_max ?
                                       std::min(sanitized_size, options.varchar_max_character_limit) :
-                                      sanitized_size);
+                                      sanitized_size) * multiplier;
```
MathMagique: Interesting question whether the additional capacity should be added before or after sanitation. Your choice is alright, but the interaction between `varchar_max_character_limit`  and `force_extra_capacity_for_unicode` should be documented somewhere, perhaps in the `advanced_usage.rst` in the `docs/pages` folder.

cpp/turbodbc/Library/src/make_description.cpp
```diff
@@ -101,10 +102,11 @@ template <typename Description>
 std::unique_ptr<description> make_character_description(cpp_odbc::column_description const & source,
                                                         turbodbc::options const & options)
 {
+    int multiplier = (options.fix_unicode_allocation ? Description::multiplier : 1);
```
MathMagique: `int const capacity_multiplier`

cpp/turbodbc/Library/turbodbc/descriptions/string_description.h
```diff
@@ -9,6 +9,7 @@ namespace turbodbc {
  */
 class string_description : public description {
 public:
+	const static int multiplier = 4;
```
MathMagique: `multiplier` does not communicate the purpose of this number well enough. I would suggest `max_code_units_per_unicode_code_point` is longer, but makes it clear that this is about Unicode and encoding (at least to those familiar with Unicode). See https://stackoverflow.com/a/27331885

cpp/turbodbc/Test/tests/make_description_of_description_test.cpp
```diff
@@ -225,6 +225,40 @@ TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsCanBeLimited)
...
 }
 
+TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsModifiedMemoryAllocation)
+{
+    std::size_t const reported_size = 10;
+    std::size_t const allocated_size = 40;
+    cpp_odbc::column_description column_description = {name, SQL_VARCHAR, reported_size, 0, supports_null_values};
+
+    turbodbc::options options;
+    options.fix_unicode_allocation = false;
```
MathMagique: Since you recycle the options struct, I think it is best to remove the names `allocation` and `modified_allocation`:
```
turbodbc::options options;

options.fix_unicode_allocation = false;
EXPECT_EQ(reported_size + 1, make_description(column_description, options)->element_size());

options.fix_unicode_allocation = true;
EXPECT_EQ(size_with_extra_capacity + 1, make_description(column_description, options)->element_size());
```

cpp/turbodbc/Test/tests/make_description_of_description_test.cpp
```diff
@@ -225,6 +225,40 @@ TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsCanBeLimited)
     EXPECT_EQ(limit + 1, limited->element_size());
 }
 
+TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsModifiedMemoryAllocation)
+{
+    std::size_t const reported_size = 10;
+    std::size_t const allocated_size = 40;
```
MathMagique: `size_with_extra_capacity` instead of `allocated_size`

cpp/turbodbc/Test/tests/make_description_of_description_test.cpp
```diff
@@ -225,6 +225,40 @@ TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsCanBeLimited)
...
+    EXPECT_EQ(reported_size + 1, allocation->element_size());
+
+    options.fix_unicode_allocation = true;
+    auto const modified_allocation = make_description(column_description, options);
+    EXPECT_EQ(allocated_size + 1, modified_allocation->element_size());
+}
+
+TEST(MakeDescriptionOfDescriptionTest, VarcharFieldsModifiedMemoryAllocationForUnicode)
+{
+    std::size_t const reported_size = 10;
```
MathMagique: comments similar to the other test apply here

python/turbodbc/options.py
```diff
@@ -52,6 +53,10 @@ def make_options(read_buffer_size=None,
      report a size of 2 billion characters.
      Please note that this option only relates to retrieving results, not sending parameters to the
      database.
+    :param fix_unicode_allocation Some ODBC drivers report the length of the varchar/nvarchar field rather
```
MathMagique: Please conform to the style of the other parameters:
1. Affects behavior/performance
2. If set to `True`, ...

cpp/turbodbc/Library/src/make_description.cpp
yaxxie: If the capacity modifier is applied before the restriction, we are no longer restricting characters but code points, which would be unexpected and lead to unexpected results, no?

MathMagique: Likewise, it could be unexpected that turbodbc requires more memory (in bytes) than the specified number of "characters" (whatever that means :-)).

But as I said, your choice is alright, and capped strings are probably more confusing than higher memory requirements. I still feel this needs externally visible documentation, though.

yaxxie: Yep, writing it up now!

MathMagique:
Changes look great, I am eagerly waiting for the build to finish and succeed :-)

yaxxie:
I see approval ... now if I can only go and figure out the other PR ;)

MathMagique:
Thanks for all the work you put in, I hope I can review your other PR soon!

PR #155, opened 2018-03-15
Adding decode utf8 from wchar parameter (related to #64) 

yaxxie:
Same remarks as #154

### DIFF
--- a/docs/pages/databases/netezza.rst
+++ b/docs/pages/databases/netezza.rst
@@ -1,0 +1,69 @@
+Netezza
+=======
+
+Although `IBM Netezza <https://www.ibm.com/analytics/netezza>`_  is not an integration tested
+database for turbodbc, some features have been added to support working with Netezza.
+Here are the recommended settings for connecting to a Netezza database via ODBC
+using the turbodbc module for Python.
+
+.. note::
+    Probably due to the prefetch buffering, turbodbc seems to be about as fast as using pyodbc,
+    however, fetching using NumPy or Arrow calls (``cursor.fetchallnumpy()`` or ``cursor.fetchallarrow()``)
+    is approximately 3 times faster due to avoiding the conversion from SQL types to Python
+    types to NumPy/Arrow types and just converting directly to NumPy/Arrow types.
+
+
+Recommended odbcinst.ini (Linux)
+--------------------------------
+
+.. code-block:: ini
+
+    [NZSQL]
+    Driver                     = /path/to/nz/lib/libnzsqlodbc3.so  # 32bit driver
+    Driver64                   = /path/to/nz/lib64/libnzodbc.so    # 64bit driver
+    UnicodeTranslationOption   = utf8
+    CharacterTranslationOption = all
+    PreFetch                   = 10000
+    Socket                     = 32000
+
+
+Recommended data source configuration
+-------------------------------------
+
+.. code-block:: ini
+
+    [Netezza]
+    Driver                     = NZSQL
+    Description                = NetezzaSQL ODBC Connection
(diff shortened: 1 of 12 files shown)


yaxxie:
Taking a look at the tests, and I'm not sure about how I can unit test this.

I thought about testing the returned type from make_description but I'm not sure on how to actually make that work. Any guidance would be appreciated.

yaxxie:
@MathMagique I'd like some input

```
    template <typename Description>
    void test_text_with_string_preference(cpp_odbc::column_description const & column_description, std::size_t expected_size)
    {
        auto const description = make_description(column_description, make_options(prefer_strings, large_decimals_as_strings));

        ASSERT_TRUE(dynamic_cast<Description const *>(description.get()))
            << "Could not convert type identifier '" << column_description.data_type << "' to expected description";

        EXPECT_EQ(expected_size, description->element_size());
        assert_custom_name_and_nullable_support(*description);
    }

    template <typename Description>
    void test_text_with_unicode_preference(cpp_odbc::column_description const & column_description, std::size_t expected_size)
    {
        auto const description = make_description(column_description, make_options(prefer_unicode, large_decimals_as_strings));

        ASSERT_TRUE(dynamic_cast<Description const *>(description.get()))
            << "Could not convert type identifier '" << column_description.data_type << "' to expected description";

        EXPECT_EQ(expected_size, description->element_size());
        assert_custom_name_and_nullable_support(*description);
    }
```

I came across that section of the test code, and if I mimic the pattern already in use, I'd need to reproduce these functions and add a few new constants instead of `prefer_unicode` and `prefer_strings` ... but I feel a bit dirty doing and I'd need to modify all the `make_options` calls, I don't mind any of that but I was wondering if you had some other preference

MathMagique:
Hi @yaxxie! I probably would not modify the `make_options()` calls, as most tests do not care about the new switch. I would probably just check what happens if I take a unicode column and default options when I change the `decode...` attribute. I'd then check for the return type with the dynamic cast. That would be two assertions for `true` and `false` in one test.

yaxxie:
How about `fetch_strings_as_utf8_varchars`?

MathMagique:
Not really correct as well; the UTF-8 behavior would be driver-specific. The only thing that is sure is that fields that are declared as `WCHAR` are retrieved as `CHAR`s. So the option name could be `fetch_wchar_as_char`

yaxxie:
@MathMagique sorry for huge wait, I finally got round to "finishing" this -- hopefully this is acceptable

yaxxie:
@MathMagique did you have a chance to review?

MathMagique:
Hey @yaxxie! Sorry, not yet :-( I will review as soon as I can.

docs/pages/databases/netezza.rst
```diff
@@ -0,0 +1,69 @@
...
+this, but this will cause turbodbc to use a lot more memory when retrieving result
+sets containing either ``VARCHAR`` or ``NVARCHAR`` fields (4x more per ``VARCHAR``
+field and 2x more per ``NVARCHAR`` field).
+
+See the :ref:`advanced options <advanced_usage_options>` for details.
+
+::
+
+    >>> from turbodbc import connect, make_options, Megabytes
+    >>> options = make_options(read_buffer_size=Megabytes(100))
```
MathMagique: I'm not sure about the buffer size, it seems to contradict the `PreFetch` option in the driver settings. Have you tried tuning the value a little towards smaller values? (This is nothing to fix, just a question/remark out of interest)

MathMagique:
Hey @yaxxie! A big thanks for a brilliant rework, and a big *sorry* that it took me so long to review your work!

PR #156, opened 2018-03-21
Replace deprecated arrow functions

pacman82:


xhochy:
We may need to adjust the setup.py

MathMagique:
You mean to generalize the requirement `'arrow': ['pyarrow>=0.8,<0.9']` to `'arrow': ['pyarrow>=0.8']`?

MathMagique:
Plus an entry in the changelog would be great

PR #167, opened 2018-05-17
Render README file on Pypi

HolgerPeters:
I realized there is no description for turbodbc on [PyPI](https://pypi.org/project/turbodbc/). This PR adds the contents of the README.md file as a setuptools `long_description` and sets the appropriate content type (see https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi). I haven't tested this with turbodbc, but I was able to sucessfully upload markdown descriptions  this way in the past.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -199,9 +199,16 @@ def get_extension_modules():
     return extension_modules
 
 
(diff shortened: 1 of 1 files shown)


MathMagique:
Thanks, Holger!

PR #171, opened 2018-06-16
turbodbc_arrow: Support more integer types for parameters

### LINKED ISSUE #170: Support more integer types in executemanycolumns(pyarrow.Table)
Currently, we only support int64 columns but users may also supply other types. As a workaround the user should cast them to an int64 column using `table.column(1).cast(pa.int64())`. Instead of the user doing this, we should accept them in https://github.com/blue-yonder/turbodbc/blob/master/cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp#L281-L319

xhochy:
Fixes #170


xhochy:
@MathMagique Added changelog and documentation. What do you think about adding a PR template like https://raw.githubusercontent.com/dask/dask/master/.github/PULL_REQUEST_TEMPLATE.md to remind people about that?

MathMagique:
Thanks for the PR @xhochy! Much appreciated!

Issue #133, opened 2017-10-06
adding support for generating batches of arrow tables


thenomemac:
I'm writing a CLI using turbdodbc's arrow support to dump the arrow arrays to disk as parquet files... I'm interested in contributed support to batch "K" rows of large tables this will allow for tables larger than memory. Would there be an easy path forward for adding this feature. Maybe I'm missing something but seems like there should be a fetcharrowbatch() method?

MathMagique:
Hi there! That sounds like a useful project of yours, and the feature you suggest makes perfect sense. There should be a `fetcharrowbatches()` similar to the existing `fetchnumpybatches()` that you can use for your tool.

MathMagique:
Done in #172

PR #172, opened 2018-06-22
Implementation for fetcharrowbatches

### LINKED ISSUE #168: Fetching as Arrow in batches
First off, turbodbc is fantastic 👍, thanks for building it!

Do you have any plans to implement fetching data as Arrow in batches?

My use case is similar to https://github.com/blue-yonder/turbodbc/issues/159, I want to fetch data and write out parquet, and holding the entire dataset in memory is inconvenient. So it would be really nice to have an Arrow equivalent of `fetchnumpybatches`.

mariusvniekerk:
Potentially closes #168 and #133

### DIFF
--- a/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
+++ b/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
@@ -171,7 +171,49 @@ Status append_to_string_builder(size_t rows_in_batch, std::unique_ptr<ArrayBuild
 }
 
 
-Status arrow_result_set::fetch_all_native(std::shared_ptr<arrow::Table>* out, bool batch_only)
+Status arrow_result_set::process_batch(size_t rows_in_batch, std::vector<std::unique_ptr<ArrayBuilder>> const& columns) {
+    // TODO: Use a PoolBuffer for this and only allocate it once
+    auto const column_info = base_result_.get_column_info();
+    auto const n_columns = column_info.size();
+    std::vector<std::reference_wrapper<cpp_odbc::multi_value_buffer const>> const buffers = base_result_.get_buffers();
+
+    std::vector<uint8_t> valid_bytes(rows_in_batch);
+    for (size_t i = 0; i != n_columns; ++i) {
+        auto const indicator_pointer = buffers[i].get().indicator_pointer();
+        for (size_t element = 0; element != rows_in_batch; ++element) {
+            if (indicator_pointer[element] == SQL_NULL_DATA) {
+                valid_bytes[element] = 0;
+            } else {
+                valid_bytes[element] = 1;
+            }
+        }
+        switch (column_info[i].type) {
+            case turbodbc::type_code::floating_point:
+                ARROW_RETURN_NOT_OK(append_to_double_builder(rows_in_batch, columns[i], buffers[i].get(), valid_bytes.data()));
+                break;
+            case turbodbc::type_code::integer:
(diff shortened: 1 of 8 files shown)


mariusvniekerk:
Probably just better names?  This does work just fine though

xhochy:
I'm fine with this PR but @MathMagique will probably have some comments.

MathMagique:
I'll try to have a look this evening (also at your PR, @xhochy)

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -224,6 +225,10 @@ Status arrow_result_set::fetch_all_native(std::shared_ptr<arrow::Table>* out)
                     break;
             }
         }
+        // Exit early if we are fetching one batch
+        if (batch_only) {
```
MathMagique: Personally, I always have to think thrice what happens whenever I read a `break` statement outside of switch cases, and the comment above indicates that the code is probably not sufficiently clear itself. May I suggest the following:

* Extract what is in the do/while loop to a helper function, with the exception of the newly added `if` and the `fetch_next_batch()` call.
* Surround the loop with an if statement:
```
if (batch_only) {
    do {
        process_batch(...);
        rows_in_batch = base_result_.fetch_next_batch();
    } while (...);
} else {
    process_batch(...);
}
```

That would make the difference between batched and non-batched mode more obvious and make the overall function smaller by moving details to `process_batch(...)`.

cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
```diff
@@ -242,10 +247,23 @@ Status arrow_result_set::fetch_all_native(std::shared_ptr<arrow::Table>* out)
     return Status::OK();
 }
 
+
+pybind11::object arrow_result_set::fetch_next_batch()
+{
+    std::shared_ptr<arrow::Table> table;
+    if (not fetch_all_native(&table, true).ok()) {
```
MathMagique: Constants `fetch_single_batch = true` and `fetch_all_batches = false` could be used here and below to clear up the meaning of `true` and `false`. If you find a way to avoid the switch in the call (e.g. by composing fetch all in terms of fetch single batch), that would be even better.

cpp/turbodbc_arrow/Library/turbodbc_arrow/arrow_result_set.h
```diff
@@ -32,7 +32,13 @@ class PYBIND11_EXPORT arrow_result_set {
      * @brief Retrieve a native (C++) Arrow Table which contains
      *        values and masks for all data
      */
-    arrow::Status fetch_all_native(std::shared_ptr<arrow::Table>* out);
+    arrow::Status fetch_all_native(std::shared_ptr<arrow::Table>* out, bool batch_only);
+
+    /**
+      * @brief Retrieve a boost Python object which contains
```
MathMagique: not a boost Python object ;-). Probably some copy/paste thing.

MathMagique:
@mariusvniekerk Thanks a lot for the efforts that you have already put in! Besides the nitpicking that I did, I think your PR is nearly there, and people will love it :-).

mariusvniekerk:
@MathMagique probably good enough now

MathMagique:
@mariusvniekerk Thanks for the changes! Looking sufficiently good now :-)

Again, thanks a lot for a great contribution. I will probably make a new release mid of next week.

ah-:
Oh great, thanks a lot for implementing this!

PR #180, opened 2018-09-30
Add Python 3.7 to AppVeyor matrix

xhochy:
Let's see if something breaks…

### DIFF
--- a/python/turbodbc/cursor.py
+++ b/python/turbodbc/cursor.py
@@ -303,7 +303,7 @@ class Cursor(object):
             result_batch = _make_masked_arrays(numpy_result_set.fetch_next_batch())
             is_empty_batch = (len(result_batch[0]) == 0)
             if is_empty_batch and not first_run:
-                raise StopIteration # Let us return a typed result set at least once
+                return # Let us return a typed result set at least once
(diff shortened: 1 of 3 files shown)


xhochy:
I have skipped OSX and Linux as they require pyarrow and its 3.7 release is still 1-2 weeks away.

python/turbodbc/cursor.py
```diff
@@ -303,7 +303,7 @@ def _numpy_batch_generator(self):
             result_batch = _make_masked_arrays(numpy_result_set.fetch_next_batch())
             is_empty_batch = (len(result_batch[0]) == 0)
             if is_empty_batch and not first_run:
-                raise StopIteration # Let us return a typed result set at least once
+                return # Let us return a typed result set at least once
```
MathMagique: Since you write "conform to PEP-479" in the change log, I would have expected a new test that would have broken with the previous code

xhochy: The code already failed on Python 3.7 in a lot of tests thus I have not added a new one.

MathMagique: Good to know, that's sufficient, then.

PR #181, opened 2018-10-10
Python 3.7 support with Arrow compatibility

xhochy:


xhochy:
Arrow Python 3.7 wheels are still missing and a changelog entry is required.

@MathMagique Should I bump to 2.8 as I have changed the minimal NumPy and Arrow versions? (1.14 is required for Python 3.7).

MathMagique:
That would even be a bump to version 3.0.0 following semantic versioning...

xhochy:
I will also kill Python 3.4 in this branch. As with many other packages Arrow has dropped the support for it in the recent release.

xhochy:
And in return, the MySQL ODBC driver is no longer maintained on Ubuntu and this not anymore available for xenial. 

@MathMagique not sure what to do here. It seems one can download the ODBC driver from the mysql homepage but not programmatically. So it's either skipping MySQL in the tests or finding another solution on how to get the driver.

MathMagique:
@xhochy Is it possible to download the MariaDB ODBC driver programatically? It should be wire-compatible with the MyODBC driver...

xhochy:
@MathMagique Finally tracked down the issue with the build in xenial. It's the same problem we were facing already with Fedora users here in the bug tracker: Using `pyarrow` manylinux1 wheels is leading to weird breakages in systems which are mostly build with the new CXX ABI. 

Given that people are also interested in `manylinux1` packages for turbodbc, I think, I'll spent a bit on creating an environment for them to build.

dhirschfeld:
Just checking - does the just released version `3.0.0` *not* support `arrow` `0.11.0`?

MathMagique:
@dhirschfeld Hi Dave! The 3.0.0 release supports arrow 0.11.x as per the requirements (https://github.com/blue-yonder/turbodbc/pull/183):

    ...
    'arrow': ['pyarrow>=0.11,<0.12'],
    ...

MathMagique:
Then the release should work for you 👍

PR #183, opened 2018-10-28
Apache Arrow 0.11.1

xhochy:
This adds support for `pyarrow==0.11.x` but does not yet introduce Python 3.7 support. I'll continue with that in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ before_install: |
   fi
 
 install:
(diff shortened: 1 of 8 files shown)


MathMagique:
The dropped Python 3.4 support needs to be reflected in the documentation and in the Readme file

xhochy:
@MathMagique Done.

PR #196, opened 2019-01-20
Use Xcode 9.4

xhochy:
6.4 is quite outdated and will be decommissioned soon. Use 9.4 as this is the same base image as conda-forge is using.


MathMagique:
Thanks @xhochy!

PR #197, opened 2019-01-24
Update for pyarrow 0.12

xhochy:


xhochy:
The things that we workaround here are fixed in https://github.com/apache/arrow/pull/3488 and https://github.com/apache/arrow/pull/3483. Thus with the next Arrow release, we can remove the workarounds again.

setup.py
```diff
@@ -216,7 +216,7 @@ def get_extension_modules():
       packages = ['turbodbc'],
       extras_require={
           # We pin Apache Arrow quite restrictively until they guarantee a stable API
-          'arrow': ['pyarrow>=0.11,<0.12'],
+          'arrow': ['pyarrow>=0.11,<0.13'],
```
MathMagique: @xhochy Would this still be compatible with pyarrow 0.11? If not, the requirements need to be changed

MathMagique:
Thanks @xhochy

PR #198, opened 2019-01-31
Add statement::finalize method and use it

### LINKED ISSUE #153: cursor.close() does not raise possible ODBC errors
```python
cursor = connection.cursor()
cursor.executemanycolumns(...)  # the data here may violate DB constraints

cursor.close()
# internal implementation:
#     cursor.impl = None
# Python destructs the inner cursor impl and prints to stdout:
#     Error while freeing statement handle: ODBC error
#     state: 27003
#     native error code: -1
#     message: [EXASOL][EXASolution driver]constraint violation - foreign key (...) (Session: ...)
#
# Instead, it should raise an exception either here or:
connection.commit()
# ^ that should raise
```

byjott:
Fixes #153

### DIFF
--- a/cpp/cpp_odbc/Library/cpp_odbc/statement.h
+++ b/cpp/cpp_odbc/Library/cpp_odbc/statement.h
@@ -155,6 +155,14 @@ public:
     bool more_results() const;
 
     virtual ~statement();
+
+    /**
+     * @brief Close all odbc ressources associated with this instance
+     * 
+     * This should only be called directly prior to the destructor; it is not
+     * valid to call any other method (except the destructor) after calling finalize.
+     */
+    void finalize();
 protected:
     statement();
 private:
@@ -182,6 +190,10 @@ private:
     virtual column_description do_describe_column_wide(SQLUSMALLINT column_id) const = 0;
     virtual column_description do_describe_parameter(SQLUSMALLINT parameter_id) const = 0;
     virtual bool do_more_results() const = 0;
+
+    virtual void do_finalize() = 0;
+
+    bool finalized = false;
 };
 
 }
--- a/cpp/cpp_odbc/Library/src/statement.cpp
+++ b/cpp/cpp_odbc/Library/src/statement.cpp
@@ -116,4 +116,12 @@ bool statement::more_results() const
     return do_more_results();
 }
 
+void statement::finalize()
+{
+    if(!finalized){
+        finalized = true;
+        do_finalize();
+    }
+}
+
 }
--- a/cpp/turbodbc/Library/src/command.cpp
+++ b/cpp/turbodbc/Library/src/command.cpp
@@ -62,4 +62,9 @@ int64_t command::get_row_count()
     }
 }
 
+void command::finalize()
+{
+    const_cast<cpp_odbc::statement &>(*statement_).finalize();
+}
+
 }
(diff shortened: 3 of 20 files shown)


cpp/cpp_odbc/Library/cpp_odbc/statement.h
```diff
@@ -155,6 +155,14 @@ class statement {
...
 
     virtual ~statement();
+
+    /**
+     * @brief Close all odbc ressources associated with this instance
+     * 
+     * This should only be called directly prior to the destructor; it is not
+     * valid to call any other method (except the destructor) after calling finalize.
+     */
+    virtual void finalize();
```
xhochy: What's the reason to make this virtual? It doesn't seem to be overloaded later on.

xhochy:
@byjott Can you please add a changelog entry?

cpp/cpp_odbc/Library/src/statement.cpp
```diff
@@ -116,4 +116,12 @@ bool statement::more_results() const
     return do_more_results();
 }
 
+void statement::finalize()
+{
+    if(!finalized){
```
MathMagique: please use

    if (not finalized)

to be consistent with the rest of the code base and to be more readable.

cpp/turbodbc/Library/src/command.cpp
```diff
@@ -62,4 +62,9 @@ int64_t command::get_row_count()
     }
 }
 
+void command::finalize()
+{
+    const_cast<cpp_odbc::statement &>(*statement_).finalize();
```
MathMagique: This is really mean ;-). Please drop the const-cast here and mark the `finalize` function as `const`. The `finalized` member then must be declared `mutable`, but I prefer that to the const-cast here.

MathMagique:
I am missing an integration test that checks the high-level bug is fixed

cpp/turbodbc/Library/src/command.cpp
byjott: It's not as simple: statement::finalize is also not const, because it needs non-const access on its inner handle. So one could go "all the way const", but that would also be weird, because "finalize" certain is (semantically speaking) a non-const operation.

cpp/cpp_odbc/Library/src/statement.cpp
byjott: Will do

byjott:
> I am missing an integration test that checks the high-level bug is fixed

Please advice where that should go and what exactly it tests. I cannot test the original bug directly without exasol. One probably could mock the "finalize" method at a certain level (either turbodbc::command or cpp_odbc::statement or the level2 api's close statement handle) to raise an error and check that the intended use of turbodbc::cursor (i.e. a call to turbodbc::cursor::reset) actually throws that error. However, the details of setting up this test is beyond what simple for me.

Note that I did manually test with exasol that the bug is not longer rpesent.

byjott:
> I am missing an integration test that checks the high-level bug is fixed

Please see the commit I just added: 71375ce
Note that I could test the tests only with postgres and mysql, so the syntax of the added test fixtures for exasol and mssql where an educated guess, but not tested yet.

cpp/cpp_odbc/Library/src/statement.cpp
byjott: I actually reverted it now to the prior version using `!`, because the MSVC compiler does not like `not`

MathMagique:
thanks @byjott for your contribution, and sorry for the long wait!

PR #202, opened 2019-02-27
Update docker

jochen-ott-by:
This modernizes the config in `contrib/docker` and uses Debian 9 as base image. As database services, the official docker images from postgres, mysql, and mssql are used. See also `contrib/docker/README` file for more.

I also fixed some issues I encountered when testing this setup.

Note that it does not install / build against arrow, because of compatibility issues of the binary pyarrow wheels with local builds that I could not resolve.


contrib/docker/Dockerfile-turbodbc
```diff
@@ -0,0 +1,23 @@
...
+
+# install the mysql odbc driver:
+RUN wget https://dev.mysql.com/get/Downloads/Connector-ODBC/8.0/mysql-connector-odbc-8.0.15-linux-debian9-x86-64bit.tar.gz && \
+    echo 7e2102246b5342f4e8e0230c2bc86589  mysql-connector-odbc-8.0.15-linux-debian9-x86-64bit.tar.gz | md5sum -c && \
+    tar xzf mysql-connector-odbc-8.0.15-linux-debian9-x86-64bit.tar.gz && \
+    cp mysql-connector-odbc-8.0.15-linux-debian9-x86-64bit/lib/libmyodbc8w.so /usr/lib/x86_64-linux-gnu/odbc/libmyodbc.so && \
+    rm -rf mysql-connector-odbc-8.0.15-linux-debian9-x86-64bit*
+
+# install the mssql odbc driver:
+RUN wget https://packages.microsoft.com/debian/9/prod/pool/main/m/msodbcsql17/msodbcsql17_17.2.0.1-1_amd64.deb && \
```
xhochy: Why not use the official apt repository as we are using the CI?

contrib/docker/Dockerfile-turbodbc
```diff
@@ -0,0 +1,23 @@
...
+    echo ab73def14f99d72bcc50d6c70ee5aa71  msodbcsql17_17.2.0.1-1_amd64.deb | md5sum -c && \
+    ACCEPT_EULA=Y dpkg -i msodbcsql17_17.2.0.1-1_amd64.deb && \
+    rm msodbcsql17_17.2.0.1-1_amd64.deb
+
+
+# create a python virtualenv at /venv that contains the necessary build and test packages:
+RUN virtualenv --python=python3.5 /venv && \
```
xhochy: Add `pyarrow==0.12.1`

contrib/docker/build_and_test.sh
```diff
@@ -1,3 +1,17 @@
-#/bin/bash
+#!/bin/bash
 
-docker run --rm -t -i -v $PWD:/io turbodbc_base /io/contrib/docker/build_and_test_in_docker.sh
+# This should run in the 'turbodbc' docker service
+
+set -ex
+
+mkdir -p /turbodbc/build_docker && cd /turbodbc/build_docker
+cmake -DCMAKE_INSTALL_PREFIX=./dist -DPYBIND11_PYTHON_VERSION=3.5 ..
```
xhochy: Use `ninja` for optimal parallelisation based on the host.

```suggestion
cmake -DCMAKE_INSTALL_PREFIX=./dist -DPYBIND11_PYTHON_VERSION=3.5 -GNinja ..
```

contrib/docker/build_and_test.sh
```diff
@@ -1,3 +1,17 @@
...
+make -j4
```
xhochy:
```suggestion
ninja
```

contrib/docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+        - MYSQL_DATABASE=test_db
+      ports:
+        - "3306"
+
+  turbodbc:
+    build:
+      context: .
+      dockerfile: Dockerfile-turbodbc
+    environment:
+       - ODBCSYSINI=/turbodbc/contrib/docker
```
xhochy: Put these environment specifications in the docker file (at the end)

contrib/docker/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+       - LC_ALL=C.UTF-8
+       # set environment variables as if 'source /venv/bin/activate' had run:
+       - VIRTUAL_ENV=/venv
+       - PATH=/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+       # configure turbodbc tests to run against all of the above databases;
+       # adapt manually if needed:
+       - TURBODBC_TEST_CONFIGURATION_FILES=query_fixtures_mysql.json,query_fixtures_postgres.json,query_fixtures_mssql.json
+    volumes:
+       # bind-mount the repository root at /turbodbc
+       - ../..:/turbodbc
```
xhochy: You can add a convenience entry for building in the likes of

```suggestion
  turbodbc-build:
    build:
      context: .
      dockerfile: Dockerfile-turbodbc
    command: contrib/docker/build_and_test.sh
```

contrib/docker/odbcinst.ini
```diff
@@ -0,0 +1,15 @@
...
+Driver=/usr/lib/x86_64-linux-gnu/odbc/libmyodbc.so
+UsageCount=1
+Threading=2
+
+[PostgreSQL Driver]
+Driver=/usr/lib/x86_64-linux-gnu/odbc/psqlodbcw.so
+FileUsage=1
+Threading=2
+
+[ODBC Driver 17 for SQL Server]
```
xhochy: Don't hardcode this but use https://github.com/blue-yonder/turbodbc/blob/a4f0c18201a7c766b330c31a029d7297f822f9bd/.travis.yml#L138 to install the correct path as the path changes with each minor version.

contrib/docker/Dockerfile-turbodbc
byjott: As already commented: "Note that it does not install / build against arrow, because of compatibility issues of the binary pyarrow wheels with local builds that I could not resolve."

contrib/docker/docker-compose.yml
byjott: Makes sense for LC_ALL, VIRTUAL_ENV and PATH, but IMO ODBCSYSINI, PYTHONPATH belong to `docker-compose.yml`, because only here the bind-mount location `/turbodbc` is defined these variables refer to.

contrib/docker/odbcinst.ini
byjott: Well, right now, the path does not change, as I hardcoded the minor version. But I see your point; will add this to `Dockerfile-turbodbc`

contrib/docker/docker-compose.yml
byjott: I'm not sure that adds much value, given that I expect developers to use the 'turbodbc' service manually anyway, and I consider running the 'build_and_test.sh' script less overall complexity than the 'turbodbc-build' service in docker-compose.yml

contrib/docker/Dockerfile-turbodbc
byjott: I'll change it to using the apt repository

contrib/docker/Dockerfile-turbodbc
byjott: I just tried again installing `pyarrow` and adjusting cmake options to include `-DDISABLE_CXX11_ABI=ON` and the tests just segfault. I can still leave it there, although my preference would be a setup in which tests actually do work.

What do you think?

xhochy: The tests should work. Leave a TODO here then and I can have a look later what's going wrong

contrib/docker/docker-compose.yml
xhochy: This is how I'm used to utilise docker things. I never login, I only run a single command in the container. I can add it later when I try to use it.

xhochy:
It would be really nice when we could also use this as the basis for the CI. Currently that is failing and it's hard to reproduce locally.

PR #205, opened 2019-03-19
Support user-provided gmock/gtest

xhochy:
Building in a conda environment where gtest and gmock are installed is
currently causing weird linking issues. Instead of failing, we should
just use the one from the environment and don't build our own.

Sadly CMake's built-in `FindGTest.cmake` doesn't support GMock so I've
bundled the modified version from Apache Arrow.


MathMagique:
Please add a Changelog entry, other than that, no objections :-)

xhochy:
> Please add a Changelog entry, other than that, no objections :-)

Done

MathMagique:
Thanks a lot, Uwe!

PR #206, opened 2019-03-21
Prepare for Arrow 0.13.0 release

xhochy:
See also https://github.com/apache/arrow/pull/4004 where I try to get this automated that we don't break it so often anymore.


xhochy:
This is ready for review, also see https://github.com/apache/arrow/pull/4004

MathMagique:
The requirements in `setup.py` would require adjustment, I believe

Issue #207, opened 2019-03-28
Parameter to truncate datetime  to 1400-9999 range 


JorgeGarciaIrazabal:
I am getting this error when trying to fetch arrow batches with timestamps that are before the year 1400. 
```
File "/opt/env/lib/python3.6/site-packages/turbodbc/cursor.py", line 335, in fetcharrowbatches
    table = rs.fetch_next_batch()
IndexError: Year is out of valid range: 1400..9999
```

Do you guys know a workaround that can help here? if not, would it be worth it to add a flag to `make_options` to truncate the dates to the valid range?

thanks in advance

xhochy:
This is a limitation in `turbodbc::date_to_days` which uses `boost::gregorian::date` internally and only supports the year range `1400..9999`. We should look for an alternative implementation that supports larger year spans.

mariusvniekerk:
@xhochy would it be possible to use the `arrow::util::date` classes in here directly?  I took a look at the code, but it does not seem obvious if anything is actually exposed to downstream libraries

PR #209, opened 2019-04-04
flag `truncate_timestamps` added to fetcharrow functions

JorgeGarciaIrazabal:
#207 
Arrow only allows timestamps with years between 1400 to 9999.
Adding the new boolean parameter `truncate_timestamps` to functions `fetchallarrow` and `fetcharrowbatches` to truncate the value to the valid range instead of throwing.


JorgeGarciaIrazabal:
It looks like OSX if failing but I don't think it is related to this PR (I might be wrong though)

MathMagique:
Hi @JorgeGarciaIrazabal - Could you please try rebasing to master and reopen the PR? OSX is fixed there.

JorgeGarciaIrazabal:
> Hi @JorgeGarciaIrazabal - Could you please try rebasing to master and reopen the PR? OSX is fixed there.

I have a different error after rebasing, I am not 100% sure why it did failed this time though.

JorgeGarciaIrazabal:
@MathMagique running the job again, all the checks passed! let me know if you want me to change/add anything to the PR. Thanks in advance

xhochy:
This is not an issue with `Arrow` but with `boost::gregorian::date` in the `date_to_days` function. I would rather fix that than have this truncation workaround.

JorgeGarciaIrazabal:
> This is not an issue with `Arrow` but with `boost::gregorian::date` in the `date_to_days` function. I would rather fix that than have this truncation workaround.

Yeah, that makes more sense. Unfortunately, with my limited c++ experience, I can not think of a better alternative for the library. Are you guys planing on working on this any time soon?

mariusvniekerk:
@wesm, @xhochy  what would be the approrpriate changes to make in Arrow,  I took at look at the datetime parts of arrow-cpp but it wasn't clear to me what we can use instead of boost::gregorian ?

PR #214, opened 2019-06-04
Follow Apache Arrow API changes for variable dictionaries coming in 0.14.x

wesm:
As a result of ARROW-3144 the dictionary values are now a member of the Array object rather than the Type object, so that the values can vary from array to array while still having semantically the same type, e.g. "dictionary-encoded string".

I don't think this should impact turbodbc too much but if you can assist with more exhaustive testing I will be happy to help sort out any issues that come up.


wesm:
cc @xhochy

wesm:
@MathMagique @xhochy how do you want to proceed on this? Are you going to wait to merge this until pyarrow 0.14.0 is available? I'll verify this branch against https://github.com/apache/arrow/pull/4577 in the meantime

wesm:
Close in favor of #219

PR #226, opened 2019-09-11
Update for Apache Arrow >= 0.15 with arrow::Column removed

wesm:
The `arrow::Column` C++ class was removed from the project. See also ARROW-6428

### DIFF
--- a/cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
+++ b/cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
@@ -351,7 +351,7 @@ namespace {
         std::vector<std::unique_ptr<parameter_converter>> converters;
 
         for (int64_t i = 0; i < table.num_columns(); ++i) {
-#if ARROW_VERSION_MAJOR > 0 || ARROW_VERSION_MINOR >= 14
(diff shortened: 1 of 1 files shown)


xhochy:
@wesm Could you try to use `#if` macros to detect the Arrow version? Then we could merge this to master and avoid https://github.com/apache/arrow/pull/5354

wesm:
Done

cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -351,7 +351,11 @@ namespace {
         std::vector<std::unique_ptr<parameter_converter>> converters;
 
         for (int64_t i = 0; i < table.num_columns(); ++i) {
-            std::shared_ptr<ChunkedArray> const & data = table.column(i)->data();
+#if ARROW_VERSION_MAJOR > 0 || ARROW_VERSION_MINOR >= 14
```
xhochy: This should be `15`?
```suggestion
#if ARROW_VERSION_MAJOR > 0 || ARROW_VERSION_MINOR >= 15
```

wesm: Yes, thank

PR #228, opened 2019-10-14
Don't destroy existing CMAKE_FLAGS environment variables if already set for unix-likes

mariusvniekerk:
As part of building a conda-build recipe that doesn't rely on the sdist on pypi, I found
that the CMAKE_CXX_FLAGS cannot be appended to at build time.  This restores that option.

### DIFF
--- a/cmake_scripts/default_build_setup.cmake
+++ b/cmake_scripts/default_build_setup.cmake
@@ -28,7 +28,7 @@ option(DISABLE_CXX11_ABI
 
 if (UNIX)
     # flags apply for both Linux and OSX!
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra $")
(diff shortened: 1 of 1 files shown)


cmake_scripts/default_build_setup.cmake
```diff
@@ -28,7 +28,7 @@ option(DISABLE_CXX11_ABI
 
 if (UNIX)
     # flags apply for both Linux and OSX!
-    set(CMAKE_CXX_FLAGS "-Wall -Wextra")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra $")
```
xhochy: @mariusvniekerk Is the trailing `$` here intended? I guess this breaks CI.

mariusvniekerk: nope lemme fix

PR #235, opened 2019-12-10
Cursors to optionally return dict with column names

darren-mayne:
Optionally have one or all cursors return results as a dictionary with the column names as keys.

set every cursor to return dict by default

`my_connection.as_dict = True`

set one cursor to return dict by default

```
my_cursor.as_dict = True
for row in my_cursor:
    print(row["mycolumn"])

```

### DIFF
--- a/python/turbodbc/connection.py
+++ b/python/turbodbc/connection.py
@@ -14,16 +14,25 @@ class Connection(object):
     def __init__(self, impl):
         self.impl = impl
         self.cursors = WeakSet([])
+        self._as_dict = False
+
+    @property
+    def as_dict(self):
+        return self._as_dict
+
+    @as_dict.setter
+    def as_dict(self, value):
+        self._as_dict = True if value is True else False
 
     @translate_exceptions
-    def cursor(self):
+    def cursor(self, as_dict=None):
         """
         Create a new ``Cursor`` instance associated with this ``Connection``
 
         :return: A new ``Cursor`` instance
         """
         self._assert_valid()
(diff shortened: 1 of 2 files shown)


xhochy:
According to https://www.python.org/dev/peps/pep-0249/#frequently-asked-questions this API was omitted for a reason, we should thus also omit it here, too.

darren-mayne:
> According to https://www.python.org/dev/peps/pep-0249/#frequently-asked-questions this API was omitted for a reason, we should thus also omit it here, too.

Unfortunate situation of lowest common denominator. Does it make a difference the default is off and this is optional?

xhochy:
I would rather provide a separate API that returns the dicts. What about `for row in my_cursor.iterrow_dicts()` ? And `fetchonedict`?

darren-mayne:
That can work too, I’ll adjust to whatever is preferred. 

One library I used called pymysql allowed the  choice of output types over the same APIs. In turbodbc’s case, they would have something like an enum property such as as_type instead of as_dict, then allowed values such as None, Dict, Numpy, Arrow. Just another approach. 

Which is preferred?

xhochy:
I prefer APIs were the types don't depend on the internal state of an object, i.e. having different methods that for each return type. @fjetter @MathMagique feel free to weigh in here.

fjetter:
> I prefer APIs were the types don't depend on the internal state of an object, i.e. having different methods that for each return type.

I feel the same way. In particular we're exercising the principle already for arrow and numpy

PR #252, opened 2020-04-12
Fixes for Arrow 0.17; add Arrow nightlies job

### LINKED ISSUE #250: turbodbc doesn't properly propagate Arrow errors
When e.g. `arrow_result_set::fetch_all` fails, it overwrites the original Arrow Status message with some generic message. It makes it harder to understand the cause of bugs or regressions. Here is an example diff that improves things:
```diff
diff --git a/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp b/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
index c2ffd17..ff23c4f 100644
--- a/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
+++ b/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp
@@ -12,6 +12,7 @@
 #include <turbodbc/time_helpers.h>
 
 #include <ciso646>
+#include <iostream>
 #include <vector>
 
 #include <boost/locale.hpp>
@@ -377,12 +378,17 @@ pybind11::object arrow_result_set::fetch_all()
     std::shared_ptr<arrow::Table> table;
     {
         pybind11::gil_scoped_release release;
-        if (not fetch_all_native(&table, false).ok()) {
-            throw turbodbc::interface_error("Fetching Arrow result set failed.");
+        auto st = fetch_all_native(&table, false);
+        if (!st.ok()) {
+            throw turbodbc::interface_error("Fetching Arrow result set failed. \n" + st.ToString());
         }
     }
     arrow::py::import_pyarrow();
-    return pybind11::reinterpret_steal<pybind11::object>(pybind11::handle(arrow::py::wrap_table(table)));
+    auto h = pybind11::handle(arrow::py::wrap_table(table));
+    if (!h) {
+        throw pybind11::error_already_set();
+    }
+    return pybind11::reinterpret_steal<pybind11::object>(h);
 }
 
```

@xhochy

### LINKED ISSUE #251: Obsolete code breaks with Arrow master
These lines:
https://github.com/blue-yonder/turbodbc/blob/master/cpp/turbodbc_arrow/Library/src/arrow_result_set.cpp#L72-L92
have long been unnecessary ([ARROW-4367](https://issues.apache.org/jira/browse/ARROW-4367) was fixed in Arrow 0.13.0), and they implement a workaround that fails validating in Arrow git master (future 0.17.0).

cc @xhochy

xhochy:
Fixes #251 
Fixes #250

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,12 @@ Version history / changelog
 
 From version 2.0.0, turbodbc adapts semantic versioning.
 
+Version 3.4.0
+-------------
+
+* Support Arrow 0.16.x and 0.17.x, require at least 0.15.x
+* Minimal NumPy version was bumped to 1.6
+
 Version 3.3.0
 -------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Version 3.4.0
 -------------
 
 * Support Arrow 0.16.x and 0.17.x, require at least 0.15.x
-* Minimal NumPy version was bumped to 1.6
+* Minimal NumPy version was bumped to 1.16
 
 Version 3.3.0
 -------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,14 @@ language: python
 jobs:
   include:
   - os: osx
-    name: CPython2.7 Arrow 0.13.0
+    name: CPython2.7 Arrow 0.15.1
     osx_image: xcode9.4
     language: generic
     env:
     - TRAVIS_PYTHON_VERSION="2.7.13"
     - TURBODBC_TEST_CONFIGURATION_FILES="query_fixtures_postgresql.json"
     - ODBC_DIR=odbc_osx
-    - TURBODBC_ARROW_VERSION=0.13.0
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -3,6 +3,12 @@ Version history / changelog
 
 From version 2.0.0, turbodbc adapts semantic versioning.
 
+Version 3.4.0
+-------------
+
+* Support Arrow 0.16.x and 0.17.x, require at least 0.15.x
+* Minimal NumPy version was bumped to 1.6
```
fjetter, line 10: Typo: This should be `1.16`

What's driving the minimal version of numpy here?

xhochy: This version is also supported for Python 3.8, 1.14 isn't. Supporting 1.14 for olders would be possible but some more bash trickery but I'd rather follow NEP-29 for NumPy versions too instead of doing extra work to support old NumPy versions.

```suggestion
* Minimal NumPy version was bumped to 1.16
```

fjetter: According to NEP-29 we should still stay on 1.15 until July but that's fine for me.

PR #269, opened 2020-06-12
doc tweaks for BOOST_ROOT and supported datatypes

keitherskine:
A couple of clarifications in the docs.  Firstly, clarify what directory the BOOST_ROOT environment variable should be pointing to.  Secondly, make it clear what database types are not supported by turbodbc.


xhochy:
Thanks @keitherskine !

PR #291, opened 2021-02-06
Introduce earthly for local and CI builds on Github Actions

bjoernmeier:
Hi,
with this PR I want to switch to Github Actions as CI as travis is going to be closed.
As it was hard to reproduce build locally I want to introduce earthly as build tool to run the builds and tests on the CI the same way as run them locally. It did really help me.
What currently is missing and I would add later are tests on Windows and OS X. I hope we could start with the current matrix to enable builds again.
What I would drop is tests for pyarrow 0.16.x.

So please let me know what you think and what is missing or should be changed to bring this in.

PR #292, opened 2021-02-08
Introduce earthly for build and dev

bjoernmeier:
Continue https://github.com/blue-yonder/turbodbc/pull/291 here to see the actions actually run as new actions are not run on fork PRs.


bjoernmeier:
I have no idea why appveyor is still here. Anyone an idea? Does it come from master?

fjetter:
The repo configuration usually includes a list of required checks for a PR merge. It is probably still configured there and we might need to adjust this afterwards. appveyor itself is probably configured to react to every push regardless of configuration

fjetter:
FYI I noticed logs pointing us to an official earthly image since docker is not pre-installed. I assume this conflicts with the mamba image
```
               +test | Docker Engine was missing. It has been installed automatically by Earthly.
               +test | Docker version 20.10.3, build 46229ca
               +test | For better use of cache, try using the official earthly/dind image for WITH DOCKER.
               +test | Docker Compose is missing. Attempting to install automatically.
               +test | Detected architecture is x86_64
               +test |   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
               +test |                                  Dload  Upload   Total   Spent    Left  Speed
               +test | 100   633  100   633    0     0   3028      0 --:--:-- --:--:-- --:--:--  3028
               +test | 100 11.6M  100 11.6M    0     0  1508k      0  0:00:07  0:00:07 --:--:-- 1644k
               +test | Docker Compose was missing. It has been installed automatically by Earthly.
               +test | docker-compose version 1.27.4, build 40524192
               +test | For better use of cache, try using the official earthly/dind image for WITH DOCKER.
```

Earthfile
```diff
@@ -0,0 +1,216 @@
...
+    ARG ARROW_VERSION_RULE=""
+    ARG NUMPY_VERSION_RULE=""
+    ARG CONDA_EXTRA=""
+    FROM --build-arg PYTHON_VERSION="$PYTHON_VERSION" \
+        --build-arg ARROW_VERSION_RULE="$ARROW_VERSION_RULE" \
+        --build-arg NUMPY_VERSION_RULE="$NUMPY_VERSION_RULE" \
+        --build-arg CONDA_EXTRA="$CONDA_EXTRA" \
+        +build
+
+    WITH DOCKER --compose ../earthly/docker-compose.yml
```
xhochy, line 113: As this is the problematic line regarding the missing `docker-compose` in the image, can someone explain to me why this needs to run in the tested image and doesn't run outside of the image as the rest of earthly seems to do?

fjetter:
I just ran ` earthly --allow-privileged +test` twice locally since this is the first command the contribution page mentions. The `+build` is not cached at all. Is this intended? (no changes in source)
I would've expected the build step to be skipped in this case

Earthfile
```diff
@@ -0,0 +1,216 @@
...
+    FROM condaforge/mambaforge:latest
+
+    ENV DEBIAN_FRONTEND=noninteractive
+    # set locale to utf-8, which is required for some odbc drivers (mysql);
+    ENV LC_ALL=C.UTF-8
+
+    RUN apt-get update && apt-get upgrade -y && \
+        apt-get install -y build-essential zlib1g-dev \
+        wget unixodbc cmake g++ \
+        odbc-postgresql postgresql-client gnupg apt-transport-https && \
```
xhochy, line 11: This should be sufficient.
```suggestion
        apt-get install -y odbc-postgresql postgresql-client gnupg apt-transport-https && \
```

fjetter:
With the non-cached `+build` and the docker-compose as part of the test, it seems I'm ending up needing to pull the docker layers, e.g. for `mysql` as well for every test run.

Apart from the caching issue, the experience is quite nice 👏

Earthfile
```diff
@@ -0,0 +1,216 @@
...
+    ARG ARROW_VERSION_RULE="<2.0.0"
+    ARG NUMPY_VERSION_RULE=""
+    ARG CONDA_EXTRA=""
+    FROM --build-arg PYTHON_VERSION="$PYTHON_VERSION" \
+        --build-arg ARROW_VERSION_RULE="$ARROW_VERSION_RULE" \
+        --build-arg NUMPY_VERSION_RULE="$NUMPY_VERSION_RULE" \
+        --build-arg CONDA_EXTRA="$CONDA_EXTRA" \
+        +python
+
+    COPY . /src
```
fjetter, line 81: This currently invalidates my local caches on every command execution, regardless of code changes or not. Haven't found the issue, yet, but I assume there is some log generated which invalidates this layer.

I assume that incremental builds are only possible if I use the interactive / manual setup you described

xhochy: Adding the offending file to `.dockerignore` should fix this?

fjetter: Is there any easy way to find the offender? Any docker build flag which prints info about the cache invalidation?

fjetter: I'm wondering what's best practice here. On could also think about a whitelisting with the dockerignore, e.g.

```
*
!cpp/**/*
!python/**/*
...
```

Earthfile
bjoernmeier: I think they decided to do this inside the container as well as you have full control over the setup there. If outside you would be dependent on docker-compose to be available externally for example. Here they install docker-compose when needed.
This of course leads to the caching issue.

One thing I had during development was a `WITH DOCKER` step where the images where pulled before doing a `COPY` with the source. So the pulled images were part of the cached steps. This is only a workaround.

Maybe we can improve it with https://docs.earthly.dev/guides/shared-cache or with https://docs.earthly.dev/guides/advanced-local-caching. Would need more time to check this. As I am a bit short of time today and tomorrow I would suggest to open an issue to improve this afterwards.

bjoernmeier:
Does anyone have an idea why the python 3.6 tests fail now after using the mambaforge images?
https://github.com/blue-yonder/turbodbc/pull/292/checks?check_run_id=1854505622

`E   ModuleNotFoundError: No module named 'turbodbc_intern'` looks a bit weird to me and I am currently out of ideas what leads to this.

Earthfile
bjoernmeier: I have to check if using a properly defined .earthignore https://docs.earthly.dev/earthfile/earthignore can improve this.

fjetter: Looks like ordinary `.dockerignore` doesn't work but we'll need an `.earthignore` https://docs.earthly.dev/earthfile/earthignore

and the magic directory to ignore is (git is not 100% necessary but e.g. VSCode fetches regularly in a background job which manipulates a file in `.git`)

```
.git*
.tmp-earthly*
```

xhochy:
> Does anyone have an idea why the python 3.6 tests fail now after using the mambaforge images?
> https://github.com/blue-yonder/turbodbc/pull/292/checks?check_run_id=1854505622
> 
> `E ModuleNotFoundError: No module named 'turbodbc_intern'` looks a bit weird to me and I am currently out of ideas what leads to this.

Install `pip` as part of the requirements, `conda` does this automatically when you install `python`, `mamba` only if requested.

docs/pages/contributing.rst
```diff
@@ -49,7 +49,89 @@ current state.
...
+
+::
+
+    earthly -i --allow-privileged +test
+
+Run a specific test setup with an export of the build package:
+
+::
+
+    earthly -P test-python3.8-arrow3.x.x
```
jochen-ott-by, line 75:
```suggestion
    earthly -P +test-python3.8-arrow3.x.x
```

jochen-ott-by (approved):
Apart from the one typo in the docs (missing "+" sign for the earthly command), it looks good to me to get this working.

xhochy:
> > Does anyone have an idea why the python 3.6 tests fail now after using the mambaforge images?
> > https://github.com/blue-yonder/turbodbc/pull/292/checks?check_run_id=1854505622
> > `E ModuleNotFoundError: No module named 'turbodbc_intern'` looks a bit weird to me and I am currently out of ideas what leads to this.
> 
> Install `pip` as part of the requirements, `conda` does this automatically when you install `python`, `mamba` only if requested.

We actually don't use `pip` anywhere so this isn't the cause :(

Earthfile
```diff
@@ -0,0 +1,216 @@
...
+
+    ENV ODBCSYSINI=/src/earthly/odbc
+    ENV TURBODBC_TEST_CONFIGURATION_FILES="query_fixtures_postgresql.json,query_fixtures_mssql.json,query_fixtures_mysql.json"
+    RUN ln -s /opt/pybind11-2.6.2 /src/pybind11
+
+    RUN bash -ic " \
+        cmake -DBOOST_ROOT=$CONDA_PREFIX -DBUILD_COVERAGE=ON \
+            -DCMAKE_INSTALL_PREFIX=./dist  \
+            -DPYTHON_EXECUTABLE=`which python` \
+            -GNinja .. && \
```
xhochy, line 94: You need to add the correct escaping as otherwise the shell commands are already executed in the outer bash context. 
```suggestion
        cmake -DBOOST_ROOT=\${CONDA_PREFIX} -DBUILD_COVERAGE=ON \
            -DCMAKE_INSTALL_PREFIX=./dist  \
            -DPYTHON_EXECUTABLE=\$(which python) \
            -GNinja .. && \
```

xhochy (commented):
The build fails currently because `cmake` picks up the wrong Python version, see the in-line fix for this.

fjetter (approved):
I think there are still a few QOL issues to figure out with this but overall I'm quite happy with this change and we can follow up on these I believe

fjetter:
Thanks @xhochy for looking into the failure. That worked perfectly!

Earthfile
xhochy: @fjetter @bjoernmeier Can we get this in, too?

Earthfile
bjoernmeier: I think this is not an issue we can solve here now and also might have to see what the earthly devs do. So I think this here is not a merge blocker?

Earthfile
bjoernmeier: Sadly I think it does not help. Seems for me the COPY is always executed.
I can check with the earthly devs but I think we can live with this for now?

Earthfile
fjetter: imo, it's an improvement over the existing setup. therefore, I wouldn't consider it a blocker

xhochy (approved):
LGTM 🚀

xhochy:
One thing we need to follow-up is a build with pure Ubuntu dependencies and `manylinux` wheels. This is the more fragile landscape.

bjoernmeier:
> One thing we need to follow-up is a build with pure Ubuntu dependencies and `manylinux` wheels. This is the more fragile landscape.

Can you write an issue for this and explain it a bit there?

I will also follow up on open topics and create issues so that we can track them better.

bjoernmeier:
Btw: I contacted @MathMagique to check the repo settings and why appveyor is not going away on master.

PR #293, opened 2021-02-10
update the arrow install requirements to limit it only downwards to >=0.17.1

bjoernmeier:
the test pipelines work for arrow >=3.0.0 as well so we should allow
it for install as well

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ From version 2.0.0, turbodbc adapts semantic versioning.
 Version 4.1.2
 -------------
 
-* Update package requirements so pyarrow<2.0 can be used.
+* Update package requirements so pyarrow>0.17.1 with no upper limit can be used.
 
 Version 4.1.1
 -------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ From version 2.0.0, turbodbc adapts semantic versioning.
 Version 4.1.2
 -------------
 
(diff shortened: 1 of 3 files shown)


fjetter (approved):

setup.py
```diff
@@ -215,7 +215,7 @@ def get_extension_modules():
       author_email = 'michael.koenig@blue-yonder.com',
       packages = ['turbodbc'],
       extras_require={
-          'arrow': ['pyarrow>=0.15,<2.0'],
```
xhochy, line 218: Please put an upper limit here. Arrow breaks regularly and this will just confuse users when it doesn't work just because of an incompatible Arrow version. I would feel more comfortable doing regular releases than having to deal with all the user reports.

bjoernmeier: Done, is the `<3.1.0` fine for you?

xhochy: Yes!

CHANGELOG.rst
```diff
@@ -6,7 +6,7 @@ From version 2.0.0, turbodbc adapts semantic versioning.
 Version 4.1.2
 -------------
 
-* Update package requirements so pyarrow<2.0 can be used.
+* Update package requirements so pyarrow>0.17.1 with no upper limit can be used.
```
xhochy, line 9:
```suggestion
* Update package requirements so pyarrow>0.17.1,<3.1 can be used.
```

xhochy (approved):

PR #308, opened 2021-05-08
Update pyarrow, numpy and python versions

alippai:
Allowing pyarrow==4.0.0 and following NEP 29, increasing minimum versions:
- Numpy 1.16->1.17
- Python 3.6->3.7


xhochy (approved):
Thanks @alippai

PR #309, opened 2021-05-09
Preprare 4.2.0 release

xhochy:


alippai:
Are we using semantic versioning? I'm wondering whether it should be version 4.2.0 or 5.0.0.

xhochy:
Yes, it is but I'm always feeling like dependency updates aren't really breaking changes.

fjetter (approved):

PR #316, opened 2021-07-20
Initialize PyDateTimeAPI in module initialization.

adament:
The static initialization causes import errors on Windows with Python
3.8 and 3.9.

This fixes the import errors on Windows with newer Python versions.

### DIFF
--- a/cpp/turbodbc_python/Library/src/determine_parameter_type.cpp
+++ b/cpp/turbodbc_python/Library/src/determine_parameter_type.cpp
@@ -100,6 +100,9 @@ python_parameter_info determine_parameter_type(pybind11::handle const & value, t
     }
 
     auto ptr = value.ptr();
+    if (!PyDateTimeAPI) {
+        PyDateTime_IMPORT;
+    }
     if (PyDateTime_Check(ptr)) {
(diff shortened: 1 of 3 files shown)


fjetter:
Do we have any tests capturing this? If not, can you add one, please?

adament:
I am not sure what you want me to test? If you want to test loading the module under windows with python >=3.8 I think @xhochy tried to do that in #254.

Unfortunately my fix was incorrect for parameter sets and caused unintialized calls to the PyDateTime api, I think this is related to multithreading but I know very little about turbodbc internals. So instead I have now tried to replicate what pybind11 does and lazily call PyDateTime_IMPORT upon usage if PyDateTimeAPI is null. This works on my local machine for parameter sets.

keitherskine:
There has been a request to add a test for this PR.  It seems to me there is an existing (failing) test in the form of `pip install turbodbc` with Python 3.8 and 3.9 on Windows.  As far as I can tell, the CI pipeline does not run tests on Windows at all (although I'm happy to be corrected on that), so all Windows tests are essentially manual.

xhochy (approved):

Issue #318, opened 2021-08-10
pyarrow 5.0 support?


ldacey:
```
name: pyarrow
channels:
  - conda-forge
dependencies:
  - python=3.9
  - pyarrow=5
  - turbodbc
```

I removed all other packages and tried to narrow down what was causing conflicts when trying to update pyarrow to 5.0, and it seems like it is turbodbc. Can this be updated on conda-forge by chance?

xhochy:
Please make a PR to the feedstock on conda-forge and add 5.0 to the conda_build_config.yaml

ldacey:
Sorry, I have not tried this before. 

I checked https://github.com/conda-forge/turbodbc-feedstock but it seems like all of the commits are mostly automated. I did not see a file name conda_build_config.yaml. I see a conda-forge.yml file but it does not look like it indicates the pyarrow version, so I must be missing something.

```
build_platform:
  osx_arm64: osx_64
provider:
  win: azure
test_on_native_only: true
conda_forge_output_validation: true
```

I see an example pull request to bump pyarrow version here (updated setup.py) which I was thinking of using as a reference:
https://github.com/blue-yonder/turbodbc/pull/308 

Would I update the setup.py / CHANGELOG to include pyarrow 5.0?

ldacey:
It seems like there is a requirement somewhere for arrow-cpp >=4.0.1,<4.0.2.0a0 which prevents the pyarrow 5.0.0 installation

Is this due to some feedstock files not being updated? It looks like all of the files are updated automatically by a bot, and some reflect the 4.2.1 update.

![image](https://user-images.githubusercontent.com/12517855/130655728-ce605a16-38a2-4f30-9ef6-3187dae33a08.png)

ldacey:
Can you point me in the right direction? I am not sure which files I should modify to get rid of the arrow_cpp conflicts. Thanks!

xhochy:
Should be fixed by https://github.com/conda-forge/turbodbc-feedstock/pull/57 It took a while to get the `arrow-cpp` migration on conda-forge going.

PR #319, opened 2021-08-13
Update pyarrow version to 5.0.0

ldacey:
Addresses #318

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,15 @@ jobs:
           - +test-python3.8-arrow0.x.x
           - +test-python3.8-arrow1.x.x
           - +test-python3.8-arrow2.x.x
+          - +test-python3.8-arrow3.x.x
+          - +test-python3.8-arrow4.x.x
+          - +test-python3.8-arrow5.x.x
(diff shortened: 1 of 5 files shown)


xhochy:
Can you also add an entry to the `.github/workflows/ci.yml` that will test against pyarrow 4/5?

ldacey:
I made an attempt based on what I could see has been done historically. It seemed like I needed to update the Earthfile as well, so I added a test per version

ldacey:
Were the tests I added incorrect? The build failed.

xhochy:
> Were the tests I added incorrect? The build failed.

Looks good, seems like a versionong issue with the MS SQL driver. I can look into that next week. Otherwise, you can also try to find to new library name.

ldacey:
I did a fresh install of msodbcsql17 on my machine and saw that the driver is now 17.8.so.1 instead of 17.7.so.2

```
ls /opt/microsoft/msodbcsql17/lib64/
libmsodbcsql-17.8.so.1.1
```

I updated the earthly/odbc/odbcinst.ini file to reflect this, hopefully that will work

xhochy (approved):
Thanks

PR #333, opened 2022-01-01
Release the GIL in python_parameter_set::flush

### LINKED ISSUE #193: parametrized queries do not release GIL on execution
Parametrized queries executed in thread do not release GIL. For my case it makes it impossible to handle query execution timeouts (i can't seem to find a way to explicitly specify command timeout in turbodbc).
MWE:
```
from threading import (
    Thread
)
from time import sleep

from turbodbc import connect

def work(label, query, parameters):
    print(f'{label} | thread start')
    try:
        connection = connect(
            driver='{SQL Server}',
            server='staging-sql',
            database='test')
        cursor = connection.cursor()
        print(f'{label} | thread before exec')
        cursor.execute(sql=query, parameters=parameters)
        print(f'{label} | thread after exec')
        data = cursor.fetchallnumpy()
        print(data)
    except Exception as ex:
        print(ex)
    print(f'{label} | thread end')

def test(label, query, parameters):
    thread = Thread(target=work, args=(label, query, parameters))
    thread.start()

    print(f'{label} | before join')
    thread.join(timeout=5)
    print(f'{label} | after join')
    print(f'{label} | isalive {thread.isAlive()}')

def signal():
    while True:
        sleep(1)
        print('signal | still alive')

signal_thread = Thread(target=signal)
signal_thread.start()

sleep(2)

test(
    label='without parameters',
    query='''
        DECLARE @intFlag INT = 1
        WHILE (1=1)
        BEGIN
           SET @intFlag = 1
        END

        select 1 as data
        ''',
     parameters=[])

sleep(2)

test(
    label='with parameters',
    query='''
        DECLARE @intFlag INT = ?
        WHILE (1=1)
        BEGIN
           SET @intFlag = 1
        END

        select 1 as data
        ''',
     parameters=[1])
```
execution result:
```
signal | still alive
without parameters | thread start
signal | still alive
without parameters | thread before exec
without parameters | before join
signal | still alive
signal | still alive
signal | still alive
signal | still alive
without parameters | after join
without parameters | isalive True
signal | still alive
signal | still alive
with parameters | thread start
with parameters | before join
signal | still alive
with parameters | thread before exec
--execution just freezes there forever--
```

note that on `cursor.execute` call with parameters background `signal` thread stops executing and it is impossible to handle query execution timeout with `thread.join(timeout)` as no `.join` call is being made due to GIL being locked.

This is not the case for the first call where no parameters are present. It successfully releases GIL so both signal thread and main thread with `thread.join` call may proceed.

### LINKED ISSUE #283: GIL in cursor.execute() if using prepared statement params
I'm using a FastAPI webserver server (with uvicorn, single process), which handles non-async requests in separate threads (`loop.run_in_executor()`).

I create the turbodbc DB connections separately in each request. So if I have 6 different concurrent request 6 different connections will be used, one in each thread. This way I have 6x(1 request, 1 thread, 1 connection, 1 cursor, 1 execute() and no fetch). The problem is if I have a slow query in one of the endpoints, the `cursor.execute()` call locks up the 5 other requests too, the server becomes unresponsive until the result is returned. 

I see that this was an issue previously (https://github.com/blue-yonder/turbodbc/pull/222, https://github.com/blue-yonder/turbodbc/pull/80) but it supposed to be fixed now, shouldn't it?

smithcre:
Resolves #193, resolves #283 

When executing a parametrized query, the ``turbodbc.cursor.execute()`` method calls ``python_parameter_set.flush()`` which makes ODBC API calls without releasing the GIL. This blocks the whole python process until the call returns. In the worst case scenario, a deadlock can occur if the operation tries to acquire a database lock which is already held by another thread in the same process.

I believe it is safe to release the GIL at this point because the ``bound_parameter_set`` class does not interact with any python objects.

Fix was tested using the sample code provided in #193 (thanks @korobkov-mindbox)

### DIFF
--- a/cpp/turbodbc_python/Library/src/python_parameter_set.cpp
+++ b/cpp/turbodbc_python/Library/src/python_parameter_set.cpp
@@ -22,6 +22,10 @@ python_parameter_set::~python_parameter_set() = default;
 
 void python_parameter_set::flush()
 {
+	// release the GIL here because bound_parameter_set::execute_batch may perform
+	// a blocking ODBC call (statement::execute_prepared)
+	pybind11::gil_scoped_release release;
+
 	// bound result set handles empty parameter sets
 	parameters_.execute_batch(current_parameter_set_);
 	current_parameter_set_ = 0;


pacman82:
> I believe it is safe to release the GIL at this point because the bound_parameter_set class does not interact with any python objects.

IMHO it also needs to be checked wether any pointers (parameter pointers and target buffers) which are dereferenced by ODBC are owned by Python objects. If that is the case releasing the GIL might invalidate these pointers during execute, causing segmentation faults.

xhochy (approved):

xhochy:
> > I believe it is safe to release the GIL at this point because the bound_parameter_set class does not interact with any python objects.
> 
> 
> 
> IMHO it also needs to be checked wether any pointers (parameter pointers and target buffers) which are dereferenced by ODBC are owned by Python objects. If that is the case releasing the GIL might invalidate these pointers during execute, causing segmentation faults.

Ops, didn't see this before merge 🙁

pacman82:
Maybe it's all good, but it would be sensible to check how the parameters are converted into C++ objects. If the Python objects are copied, all good, but if they are only borrowed, than this might result in a segmenation fault.

xhochy:
Even borrowing would be ok as long as the code doesn’t do any calls to the Python C-API. The reference count will keep the objects alive.

pacman82:
> Even borrowing would be ok as long as the code doesn’t do any calls to the Python C-API. The reference count will keep the objects alive.

I would be cautios here. After all ODBC doesn't bind pointers to Python Objects, but (maybe) pointers owned by python objects. Mutating these (Python Objects) might very well invalidate the pointers, or violate invariants (such as between the length indicator and the bound parameter buffer).

Again didn't read through the binding code, and if everything is passed as a copy, no harm done. In any other case I would disadvise releasing the GIL.

smithcre:
See the parameter binding code starting here: https://github.com/blue-yonder/turbodbc/blob/3cccd0af637944cf566c6c295c92f23ba90deffa/cpp/turbodbc_python/Library/src/python_parameter_set.cpp#L64

It seems to me that the ``parameter_converter`` code in ``determine_parameter_type.cpp`` always makes a copy of the python objects that are passed in (and the GIL is held during this stage of the process).

I don't believe that the ``bound_parameter_set.execute_batch()`` method interacts with any python object pointers, so it should be safe to release the GIL before calling it.

Parameter binding code:
https://github.com/blue-yonder/turbodbc/blob/3cccd0af637944cf566c6c295c92f23ba90deffa/cpp/turbodbc_python/Library/src/determine_parameter_type.cpp

pacman82:
> It seems to me that the parameter_converter code in determine_parameter_type.cpp always makes a copy of the python objects that are passed in

All good then 👍

Issue #336, opened 2022-02-03
Add support for pyarrow 7.0.0


ldacey:
New pyarrow version was released today

alippai:
It won't be compatible with the current TurbODBC release, also it's not available on conda yet

ldacey:
Yep, it is not on conda yet. I prepared the pull request for when it is and increased the turbodbc version to 4.3.1. I am not sure if there were any breaking changes with this release though.

alippai:
Yes, there is at least one breaking change

PR #337, opened 2022-02-03
Add support for pyarrow 7.0.0

ldacey:
Addresses #336 

The newest pyarrow version is not on conda so this won't run properly yet. Hopefully good to go otherwise.


xhochy:
Triggered a re-run of CI here with the new arrow release but the jobs still fail.

alippai:
https://github.com/apache/arrow/pull/12144#issuecomment-1012396961 this is the reason

alippai:
Fixed in https://github.com/blue-yonder/turbodbc/pull/338

PR #340, opened 2022-02-13
Update 3rd parties

alippai:
This simplifies the build code and build time a little bit. The removal of numpy 1.17 and 1.18 is following the NEP 29 guide. Dropping Arrow 0.15 and 0.17 matches the Pandas (1.4) decision and allows to remove some macros too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,8 @@ jobs:
             upload: false
     steps:
       - uses: actions/checkout@v2
-      - name: Download earthly 0.5.24
-        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.5.24/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
+      - name: Download earthly 0.5.0
+        run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.5.0/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
       - name: Run build
         run: earthly --allow-privileged ${{ matrix.target }}
         continue-on-error: ${{ matrix.experimental }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,8 +62,8 @@ jobs:
             upload: false
     steps:
       - uses: actions/checkout@v2
-      - name: Download earthly 0.6.7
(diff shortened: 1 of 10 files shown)


alippai:
@xhochy we might want to include this before the conda (turbodbc feedstock) release

alippai:
@bjoernmeier @bjoern-meier-by I wanted to update earthly as well, but there are breaking changes (even on the 0.5.x series) and I'm not familiar what it does exactly - the naive upgrade didn't work. I saw there are many improvements added in 0.6, it could greatly simplify the build

xhochy:
Actually, I would like to do one release before this. Releases here are cheap.

.github/workflows/ci.yml
```diff
@@ -64,7 +62,7 @@ jobs:
             upload: false
     steps:
       - uses: actions/checkout@v2
-      - name: Download earthly 5.0.0
+      - name: Download earthly 0.5.0
```
xhochy, line 65: This should probably be reverted?

setup.py
```diff
@@ -215,8 +215,8 @@ def get_extension_modules():
       author_email = 'michael.koenig@blue-yonder.com',
       packages = ['turbodbc'],
       extras_require={
```
xhochy, line 217: Can you add a `python_requires'`>=3.8'` here?

.github/workflows/ci.yml
alippai: I think this was a typo?!

xhochy: Ah!

ldacey:
Is this still being released to conda? I see 4.4 is up for pypi. It seems like 4.3.1 will not work with the new pyarrow:

 `- package turbodbc-4.3.1-py39h8f56454_0 requires arrow-cpp >=4.0.1,<4.0.2.0a0, but none of the providers can be installed`

I only bring this up because I checked the conda-forge feedstock page and noticed the last commits reference 4.3.1 instead of 4.4.0 so I just want to make sure this is in progress and we are just waiting.

`regro-cf-autotick-bot [turbodbc v4.3.1`

xhochy:
Just merged https://github.com/conda-forge/turbodbc-feedstock/pull/64, next time feel free to ping there or even better, add yourself as a maintainer on the feedstock (open an issue with `@conda-forge-admin, please add user @ldacey`)

ldacey:
Sure, will do. I assumed that the conda build was delayed compared to pypi so I was attempting to be patient and did not check until a bit ago. Thanks a lot though, appreciate it!

Issue #354, opened 2022-05-10
Add support for pyarrow 8.0


ldacey:
Latest version of pyarrow was released recently.

PR #356, opened 2022-05-11
#354- Add support for pyarrow 8.0

alippai:


alippai:
@ldacey @xhochy now we have greens. I'm not sure about the gmock pinning, but I had no better idea how to fix the test issue. Also I didn't check the dependency version rules added by @ldacey just built on that commit as-is. If we fix the precommit error (the action can't get h the repo???) This can be merged and gmock upgraded in a separate PR

xhochy (approved):

Issue #368, opened 2022-12-01
Earthly testing failing on all combinations


david-engelmann:
[Last previous successful run - Nov 27th, 22](https://github.com/blue-yonder/turbodbc/actions/runs/3562433286)

All of the previously successful runs fail with the same message.
```
Sqlcmd: Error: Microsoft ODBC Driver 17 for SQL Server : Can't open lib '/opt/microsoft/msodbcsql17/lib64/libmsodbcsql-17.10.so.1.1' : file not found.
```
The error occurs on my local master and feature branches.

david-engelmann:
Need to change the driver source in the ./earthly/odbc/odbcinst.ini file from 

```
Driver = /opt/microsoft/msodbcsql17/lib64/libmsodbsql-17.10.so.1.1
```
to
```
Driver = /opt/microsoft/msodbcsql17/lib64/libmsodbsql-17.10.so.1.1
```

I'm adding the change to my feature branch for multi result set support. PR coming shortly

Issue #366, opened 2022-11-02
Add support for pyarrow 10.0


david-engelmann:
Latest version of pyarrow was released and is not supported by turbodbc

david-engelmann:
pyarrow10 support is possible with the merge of https://github.com/blue-yonder/turbodbc/pull/369

david-engelmann:
#369 merged

Issue #367, opened 2022-11-16
Store Procedures should return multiple result sets


david-engelmann:
Calling a store procedure that returns multiple results is not possible currently in Turbodbc, I'd like to add in this functionality. pyodbc currently has `cursor.nextset()` to accomplish this goal. [More details on the pyodbc implementation of nextset can be found here](https://github.com/mkleehammer/pyodbc/wiki/Cursor#nextset). 

[Related Turbodbc Issue #1](https://github.com/blue-yonder/turbodbc/issues/323)

[Related Turbodbc Issue #2](https://github.com/blue-yonder/turbodbc/issues/145)

[Here is a link to where it is implemented in pyodbc](https://github.com/mkleehammer/pyodbc/blob/ec1e3a78b808a713f50c9c776e691dc1edc97e5d/src/cursor.cpp#LL1891).

david-engelmann:
### Needed  cpp Functions ###
- currentDataSet
```
dataset number out of the total number of result sets
```
- dataSetCount
```
total number of result sets
```
- hasMoreDataSets
```
currentDataSet() + 1 < dataSetCount();
```
- getNextDataSet
```
call SQLMoreResults and then return Result Set
```

david-engelmann:
Ready to be closed out with the merge of #369

david-engelmann:
Closed out

Issue #323, opened 2021-10-12
Result Set from Stored Procedure Returns Only First Result


alexdavis24:
Running the following code:

```
# some params defined
cnxn = connect(driver='{ODBC Driver 17 for SQL Server}', server=srv, database=db, uid=user, pwd=password,
                            turbodbc_options=options)
crsr = cnxn.cursor()
cmd = "EXEC someDB.dbo.TEST_PROC"
crsr.execute(cmd)
result = crsr.fetchallnumpy()
print(result)
```

where `someDB.dbo.TEST_PROC`
is defined as 
```
CREATE PROC TEST_PROC
AS
SELECT 1
SELECT 2
```

Rather than getting the final result `2`, or both results, only `1` is returned.

Came across this issue when inserting into a temp table in a proc, followed by a select statement, which executed fine using SSMS. When moving to Python/turbodbc, however, I received an error that there was no available result set. Tried fetchall, fetchmany, etc. without much success. Any advice on how to work around this or explicitly pull the last result set would be greatly appreciated.

I am running ODBC Driver 17 for SQL Server, turbodbc 4.1.2 installed through conda-forge, and python 3.9.4. I've been using a turbodbc interface for a couple years, so I am moderately familiar with the package, but this behavior had me stumped!

alexdavis24:
Does returning the this violate the Principal of Least Surprise - running a query in SSMS such as 
```
select 1
create table #tmptbl(val int)
```

or flipped
```
create table #tmptbl(val int)
select 1
```
will return a result set of 1, ie, a result set is always prioritized over a row count or return message. Wouldn't we expect the same behavior to be mirrored in our ODBC connections?

david-engelmann:
This should be closed out due to #369

PR #369, opened 2022-12-01
Support Multiple Result Sets

david-engelmann:
add in support for cursor.nextset()

Fixes issue with Earthly not finding the correct mssql driver


david-engelmann:
Fixes:
https://github.com/blue-yonder/turbodbc/issues/368
https://github.com/blue-yonder/turbodbc/issues/367
https://github.com/blue-yonder/turbodbc/issues/323
https://github.com/blue-yonder/turbodbc/issues/145
https://github.com/blue-yonder/turbodbc/issues/192

david-engelmann:
Fixes: 
https://github.com/blue-yonder/turbodbc/issues/366

h-vetinari:
> Fixes: #366

How does this affect pyarrow 10 support? As long as

https://github.com/blue-yonder/turbodbc/blob/ca36325a61cda7e1af628b7a947de8a3a00149be/setup.py#L233
is there, support is explicitly disabled

david-engelmann:
I have all the commits ready to be push to this feature branch to support pyarrow10. There is one issue calling new test around pyarrow10 to fail. [Once this arrow issue is closed out, I'll commit my changes](https://github.com/apache/arrow/issues/14931)

setup.py
```diff
@@ -191,6 +191,12 @@ def get_extension_modules():
         pyarrow_location = os.path.dirname(pyarrow.__file__)
         # For now, assume that we build against bundled pyarrow releases.
         pyarrow_include_dir = os.path.join(pyarrow_location, "include")
```
kou, line 193: Should we remove this?

cmake_scripts/FindArrow.cmake
```diff
@@ -106,10 +106,10 @@ if (ARROW_INCLUDE_DIR AND ARROW_LIBS)
     set(ARROW_PYTHON_SHARED_IMP_LIB ${ARROW_LIBS}/${ARROW_PYTHON_LIB_NAME}.lib)
   else()
     set(ARROW_STATIC_LIB ${ARROW_PYTHON_LIB_PATH}/libarrow.a)
-    set(ARROW_PYTHON_STATIC_LIB ${ARROW_PYTHON_LIB_PATH}/libarrow_python.a)
+    set(ARROW_PYTHON_STATIC_LIB ${ARROW_PYTHON_LIB_PATH}/python3.8/site-packages/pyarrow/libarrow_python.a)
```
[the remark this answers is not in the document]
david-engelmann: We could but it will require running a python script from a cmake file

setup.py
```diff
@@ -199,14 +205,25 @@ def get_extension_modules():
...
         else:
             pyarrow_module_link_args.append("-Wl,-rpath,$ORIGIN/pyarrow")
+
+        try:
+            arrow_libs = pyarrow.get_libraries()
+        except:
+            arrow_libs = ["arrow", "arrow_python"]
+
+        try:
+            arrow_lib_dirs = pyarrow.get_library_dirs() + [pyarrow_location]
```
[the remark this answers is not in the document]
david-engelmann: I've updated this

setup.py
```diff
@@ -199,14 +205,25 @@ def get_extension_modules():
             pyarrow_module_link_args.append("-Wl,-rpath,@loader_path/pyarrow")
         else:
             pyarrow_module_link_args.append("-Wl,-rpath,$ORIGIN/pyarrow")
+
+        try:
+            arrow_libs = pyarrow.get_libraries()
+        except:
+            arrow_libs = ["arrow", "arrow_python"]
```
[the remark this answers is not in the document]
david-engelmann: I made these adjustments

setup.py
david-engelmann: removed in latest commit

cmake_scripts/FindArrow.cmake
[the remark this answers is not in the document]
jorisvandenbossche: But I suppose it should be possible to not hardcode the `python3.8` ? 

Above in this file, I see other places where `site-packages` is referenced without hardcoding this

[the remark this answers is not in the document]
david-engelmann: @jorisvandenbossche I made that adjustment, I'm not able to confirm the libarrow_python.a exists in that location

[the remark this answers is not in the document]
jorisvandenbossche: As a possible alternative, and to basically do what @kou suggested, it might be worth to take a look at what FindNumpy.cmake does. 

Here in turbodobc, you seem to use `find_path` with a set of options:

https://github.com/blue-yonder/turbodbc/blob/ca36325a61cda7e1af628b7a947de8a3a00149be/cmake_scripts/Findnumpy.cmake#L8

In arrow, our FindNumpy.cmake actually does run a bit of python:

https://github.com/apache/arrow/blob/b21bf749f291367f85ad61751205e7deeee92bc7/cpp/cmake_modules/FindNumPy.cmake#L49-L50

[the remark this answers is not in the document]
david-engelmann: This is perfect! I'm going to clean everything up now

cmake_scripts/FindArrow.cmake
```diff
@@ -107,9 +127,8 @@ if (ARROW_INCLUDE_DIR AND ARROW_LIBS)
   else()
     set(ARROW_STATIC_LIB ${ARROW_PYTHON_LIB_PATH}/libarrow.a)
     set(ARROW_PYTHON_STATIC_LIB ${ARROW_PYTHON_LIB_PATH}/libarrow_python.a)
```
david-engelmann, line 129: still running into the `fatal error: arrow/python/pyarrow.h: No such file or directory` issue

david-engelmann:
[Test pass for pyarrow <10, but fail for pyarrow10](https://github.com/david-engelmann/turbodbc/pull/2/checks). Seems like the STATIC_LIB files could be the issue, ie. libarrow.a and libarrow_python.a.

setup.py
```diff
@@ -230,11 +234,11 @@ def get_extension_modules():
     packages=["turbodbc"],
     setup_requires=[
         "pybind11>=2.2.0",
-        "pyarrow>=1,<9.1.0",
+        "pyarrow>=1,<10.0.2",
```
h-vetinari, line 237: Is there any reason to expect that things would break in a patch release?

Why not `<10.1` or even `<11`?

david-engelmann: I'll change it to 11, that's what it's in the test

david-engelmann:
Seems to me there are two include paths post pyarrow10, one is the old `/opt/conda/envs/turbodbc-dev/include` and one is the new `/opt/conda/envs/turbodbc-dev/lib/python*.*/site-packages/pyarrow/include` (PYARROW_INCLUDE_DIR) where things like api.h and pyarrow.h now live. additionally libarrow.a is required, here is the relevant doc below

```
Find ARROW (arrow/api.h, libarrow.a, libarrow.so)
This module defines
ARROW_INCLUDE_DIR, directory containing headers
ARROW_LIBS, directory containing arrow libraries
ARROW_STATIC_LIB, path to libarrow.a
ARROW_SHARED_LIB, path to libarrow's shared library
ARROW_FOUND, whether arrow has been found
```

jorisvandenbossche:
> Seems to me there are two include paths post pyarrow10

Yes, that is correct and expected

> additionally libarrow.a is required

Do you need the static libraries? For example, when installing pyarrow as a wheel, only .so libs are included for libarrow(_*). And certainly now libarrow_python is always included in the pyarrow package (also for conda), I think this will always only available as dynamic library.

xhochy:
We could altogether remove the static section from `FindArrow`, this is not needed for turbodbc.

david-engelmann:
> > Seems to me there are two include paths post pyarrow10
> 
> Yes, that is correct and expected
> 
> > additionally libarrow.a is required
> 
> Do you need the static libraries? For example, when installing pyarrow as a wheel, only .so libs are included for libarrow(_*). And certainly now libarrow_python is always included in the pyarrow package (also for conda), I think this will always only available as dynamic library.

I've adjusted the FindArrow logic to not include static libraries based on @xhochy comment. The two include path issue still persist for pyarrow10 test. Currently, the FindArrow logic defines the following `ARROW_INCLUDE_DIR, directory containing headers`. I'm going to define PYARROW_INCLUDE_DIR and add an include_directories call in cpp/turbodbc_arrow/CMakeLists.txt

david-engelmann:
Currently this feature branch works for all test cases including pyarrow10. The next steps is to add in the changes to the change log. The largest adjustments are the introduction of .nextset() capabilities, and the ability to use pyarrow10+. The largest overarching changing that users should be aware of is that pyarrow10 requires the use of c++17, I bumped the default arguments but that's a pretty large change we will have to make with pyarrow

cmake_scripts/FindArrow.cmake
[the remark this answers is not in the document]
david-engelmann: pyarrow.get_library_dirs() was attempted before completely removing the static references from FindArrow. pyarrow.get_include() is used for the newly created include paths post pyarrow10 changes

cmake_scripts/FindArrow.cmake
david-engelmann: resolved by adding in the PYARROW_INCLUDE_DIR and calling it in the include_directories call in cpp/turbodbc_arrow/CMakeLists.txt

cmake_scripts/FindArrow.cmake
```diff
@@ -41,19 +41,35 @@ if (NOT ARROW_HOME)
     pkg_get_variable(ARROW_SO_VERSION arrow so_version)
     message(STATUS "Arrow SO version: ${ARROW_SO_VERSION}")
     set(ARROW_INCLUDE_DIR ${ARROW_INCLUDE_DIRS})
```
david-engelmann, line 43: Should ARROW_INCLUDE_DIRS have an s on the end?

cmake_scripts/FindArrow.cmake
```diff
@@ -41,19 +41,35 @@ if (NOT ARROW_HOME)
...
+    set(PYARROW_INCLUDE_DIR ${PYARROW_INCLUDE_DIR})
```
kou, line 44: We need to use `pa_get_include()` here too.

cmake_scripts/FindArrow.cmake
```diff
@@ -41,19 +41,35 @@ if (NOT ARROW_HOME)
...
     set(ARROW_LIBS ${ARROW_LIBRARY_DIRS})
     set(ARROW_SEARCH_LIB_PATH ${ARROW_LIBRARY_DIRS})
   elseif(DEFINED ENV{VIRTUAL_ENV})
     find_path(ARROW_INCLUDE_DIR arrow/api.h HINTS
       $ENV{VIRTUAL_ENV}/lib/*/site-packages/pyarrow/include)
```
kou, line 49: We can use `PYARROW_INLCUDE_DIR` defined below for `ARROW_INCLUDE_DIR` too.

cmake_scripts/FindArrow.cmake
```diff
@@ -41,19 +41,35 @@ if (NOT ARROW_HOME)
...
+    execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c" "import pyarrow as pa; print(pa.get_include());"
+                RESULT_VARIABLE _PYARROW_SEARCH_SUCCESS
+                OUTPUT_VARIABLE PYARROW_INCLUDE_DIR
+                ERROR_VARIABLE _PYARROW_ERROR_VALUE
+                OUTPUT_STRIP_TRAILING_WHITESPACE)
     get_filename_component(ARROW_SEARCH_LIB_PATH ${ARROW_INCLUDE_DIR} DIRECTORY)
```
kou, line 55: I think that we can use this logic for `elseif(DEFINED ENV{VIRTUAL_ENV})`, `else() if (MSVC)` and `else() else()` cluases.

cmake_scripts/FindArrow.cmake
```diff
@@ -87,8 +109,10 @@ message(STATUS "Found ${ARROW_LIB_PATH} in ${ARROW_SEARCH_LIB_PATH}")
 get_filename_component(ARROW_LIBS ${ARROW_LIB_PATH} DIRECTORY)
 
 find_library(ARROW_PYTHON_LIB_PATH NAMES arrow_python
-  PATHS
-  ${ARROW_SEARCH_LIB_PATH}
+    PATHS
+    ${ARROW_SEARCH_LIB_PATH}/*/site-packages/pyarrow
```
kou, line 113: Do we still need this?

cmake_scripts/FindArrow.cmake
kou: Could you use `pa.get_library_dirs()` for `ARROW_SEARCH_LIB_PATH`?

cmake_scripts/default_build_setup.cmake
```diff
@@ -14,23 +14,23 @@ endif()
...
 # build shared instead of static libraries
 set(BUILD_SHARED_LIBS TRUE)
 
 option(BUILD_COVERAGE
        "Determines whether C++ files will be built with support for coverage measurement"
        OFF)
-option(DISABLE_CXX11_ABI
-       "Determines whether the C++ compiler shall link to a C++11 conforming standard library, see https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html for details"
+option(DISABLE_CXX17_ABI
+       "Determines whether the C++ compiler shall link to a C++17 conforming standard library, see https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html for details"
```
kou, line 26: We should not change this. It's only for C++11.

cmake_scripts/FindArrow.cmake
david-engelmann: We have this that I can switch out the get_filename_component call for

```
execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c" "import pyarrow as pa; print(pa.get_library_dirs());"
            RESULT_VARIABLE _PYARROW_SEARCH_SUCCESS
            OUTPUT_VARIABLE _PYARROW_VALUES_OUTPUT
            ERROR_VARIABLE _PYARROW_ERROR_VALUE
            OUTPUT_STRIP_TRAILING_WHITESPACE)

# convert to the path needed
string(REGEX REPLACE "," ";" _PYARROW_VALUES ${_PYARROW_VALUES_OUTPUT})
string(REGEX REPLACE "\\]" "" _PYARROW_VALUES ${_PYARROW_VALUES})
string(REGEX REPLACE "\\[" "" _PYARROW_VALUES ${_PYARROW_VALUES})
list(GET _PYARROW_VALUES 0 ARROW_SEARCH_LIB_PATH)
```

cmake_scripts/FindArrow.cmake
david-engelmann: remove the find_path call and run set(ARROW_INCLUDE_DIR ${PYARROW_INCLUDE_DIR}) after the get_include call

cmake_scripts/default_build_setup.cmake
david-engelmann: I've reverted that

xhochy (approved):
Looks good from my side, leaving this open for @kou and @jorisvandenbossche to review and will merge and release on Monday if there is no objection.

cmake_scripts/FindArrow.cmake
kou: Yes. I think so.

kou (approved):
+1

cmake_scripts/FindArrow.cmake
david-engelmann: I've made this change

PR #372, opened 2023-01-17
Improve support for multithreading

nominee1997:
Currently we are quite conservative with releasing the GIL - this lowers performance in multithreaded use and also creates possibility for deadlocks when used with connection poolers (i.e. SQL Governor / pgpool).

This PR adds releasing the GIL in various key bits of the code which perform blocking calls (using pybind11::call_guard<pybind11::gil_scoped_release>).

I've tested this in a multithreaded app and performance improvement was ~30% or so, though of course YMMV.

I think these changes should be safe - we have dbapi threadsafety set to 1, so clients expect that connections are not safe to be shared between threads.

Please let me know what you think!

### DIFF
--- a/cpp/turbodbc_python/Library/src/python_bindings/connect.cpp
+++ b/cpp/turbodbc_python/Library/src/python_bindings/connect.cpp
@@ -6,7 +6,7 @@ namespace turbodbc { namespace bindings {
 
 void for_connect(pybind11::module & module)
 {
-    module.def("connect", turbodbc::connect);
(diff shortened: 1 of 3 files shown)


nominee1997:
Any thoughts on the PR? Anything blocking from merge?

dhirschfeld:
I'm using `turbodbc` in a multi-threaded app and would be very interested in anything which increased the performance!

nominee1997:
Is this project still actively maintained? Happy to discuss the change in more detail, I don't see anything stopping from merging. All tests pass without issue. Are there any actions we could do to get this merged in?

PR #376, opened 2023-01-25
Fix g++ ABI compatibility issue with pyarrow 3.9/3.10 on Linux

### LINKED ISSUE #375: Arrow support doesn't work in 3.10 with pyarrow-9.0.0/10.0.1 on Ubuntu 22.04
Environment:
 Ubuntu Docker image tag: 22.04
 Python: 3.10.6
 pyarrow: 9.0.0 or 10.0.1
 turbodbc: 4.5.6  (pip installed after pyarrow)

```
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] on linux
>>> import turbodbc_arrow_support
ImportError: /opt/deephaven/venv/lib/python3.10/site-packages/turbodbc_arrow_support.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZNK5arrow6Status8ToStringB5cxx11Ev
>>>
```

jmao-denver:
Fixes #375

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import distutils.sysconfig
+import distutils.ccompiler
 import itertools
 import os
 import os.path
@@ -125,10 +126,16 @@ else:
     odbclib = "odbc"
 
 
+def _get_cxx_compiler():
+    cc = distutils.ccompiler.new_compiler()
+    distutils.sysconfig.customize_compiler(cc)
+    return cc.compiler_cxx[0]
+
+
 def is_cxx11_abi():
-    cxx11_abi = subprocess.getoutput("echo '#include <string>' | g++ -x c++ -E -dM - | fgrep _GLIBCXX_USE_CXX11_ABI | \
-                                     cut -d ' ' -f2,3")
-    return cxx11_abi == '_GLIBCXX_USE_CXX11_ABI 1'
+    shcmd = f"echo '#include <string>' | {_get_cxx_compiler()} -x c++ -E -dM - | fgrep _GLIBCXX_USE_CXX11_ABI |  cut -d ' ' -f2,3"
+    cxx11_abi = subprocess.getoutput(shcmd)
+    return cxx11_abi == "_GLIBCXX_USE_CXX11_ABI 1"
 
 
 def get_extension_modules():
@@ -212,7 +219,7 @@ def get_extension_modules():
             pyarrow_module_link_args.append("-Wl,-rpath,@loader_path/pyarrow")
         else:
             pyarrow_module_link_args.append("-Wl,-rpath,$ORIGIN/pyarrow")
-            if not os.environ.get('CONDA_BUILD') and is_cxx11_abi():
+            if not os.environ.get("CONDA_BUILD") and is_cxx11_abi():
                 extra_compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=0")
 
         arrow_libs = pyarrow.get_libraries()
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import distutils.sysconfig
 import itertools
 import os
 import os.path
+import subprocess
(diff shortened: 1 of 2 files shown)


xhochy (changes requested):
This will break at least conda-based installations. Please add a check to find the correct `_GLIBCXX_USE_CXX11_ABI` in your build setup

setup.py
```diff
@@ -119,6 +120,12 @@ def __str__(self):
     odbclib = "odbc"
 
 
+def is_cxx11_abi():
+    cxx11_abi = subprocess.getoutput("echo '#include <string>' | g++ -x c++ -E -dM - | fgrep _GLIBCXX_USE_CXX11_ABI | \
```
xhochy, line 124: The `g++` executable should be exchanged with the compiler detected by the build system. It isn't always called `g++` but could be sometimes `x86-linux-g++` or similar.

xhochy (changes requested):

setup.py
jmao-denver: Would changing it to gcc solve the problem?

xhochy: No, there is some variable you can get from distutils or setuptools that tell you the current CXX compiler executable.

jmao-denver: Thanks for the tip, did some digging and found a solution, please review

setup.py
```diff
@@ -1,7 +1,9 @@
 import distutils.sysconfig
+import distutils.ccompiler
```
xhochy, line 2:
```suggestion
import distutils.ccompiler
import distutils.sysconfig
```

setup.py
```diff
@@ -200,6 +214,8 @@ def get_extension_modules():
             pyarrow_module_link_args.append("-Wl,-rpath,@loader_path/pyarrow")
         else:
             pyarrow_module_link_args.append("-Wl,-rpath,$ORIGIN/pyarrow")
+            if not os.environ.get("CONDA_BUILD") and is_cxx11_abi():
```
xhochy, line 217: I'm a bit confused about the condition here. We add `ABI=0` for the case that `is_cxx11_abi` is true. Shouldn't it be the other way around?

Weirdly the tests pass :D

setup.py
```diff
@@ -119,6 +121,18 @@ def __str__(self):
     odbclib = "odbc"
 
 
+def _get_cxx_compiler():
+    cc = distutils.ccompiler.new_compiler()
+    distutils.sysconfig.customize_compiler(cc)
+    return cc.compiler_cxx[0]
```
xhochy, line 127:
```suggestion
    return cc.compiler_cxx[0]  # type: ignore
```

setup.py
jmao-denver: That's because for whatever reasons the arrow libraries built for pyarrow are built with ABI=0. So turbodbc would build but the arrow support is broken as described in #375

xhochy: Yes, `pyarrow` `manylinux` wheels are built with `ABI=0`. But you set `_GLIBCXX_USE_CXX11_ABI=0` in the case that `_GLIBCXX_USE_CXX11_ABI` is set to `1` in the compiler output. This is a bit contradicting.

jmao-denver: Yes, it is confusing and it is all because of https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html. Basically, if we compile turbodbc with _GLIBCXX_USE_CXX11_ABI set to 1, then any declaration of std::string becomes std::__cxx11:string and at runtime the linker won't be able to resolve any arrow methods with std::string in the definitions.

xhochy:
@jmao-denver I used a slightly different check that scans the binary for a known symbol that is present in all versions of `pyarrow` but different between different CXX11ABIs. I'll merge and release this now. Thanks!

xhochy (approved):

PR #382, opened 2023-02-15
Fix import of turbodbc_arrow on Windows

xhochy:


CHANGELOG.rst
```diff
@@ -3,6 +3,12 @@ Version history / changelog
 
 From version 2.0.0, turbodbc adapts semantic versioning.
 
+Version 4.5.8
+-------------
+
+* Ensure we can correctly load ``arrow_python.dll`` on Windows.
+* Various build system fixes as we now have macOS and Windows CI.
```
h-vetinari, line 10: Shouldn't it be called out that the recommended API (certainly on windows) is now `import turbodbc.arrow_support` rather than `turbodbc_arrow_support`? (dot vs. underscore)

xhochy: Actually, neither of them should be used by users directly. Users should only import `turbodbc.cursor`.

PR #386, opened 2023-03-20
create a four result test with column names

david-engelmann:


david-engelmann:
meant to create for my fork

PR #400, opened 2023-11-14
Fix slowdown from utf conversion

SimeonStoykovQC:
## Motivation

4.5.8 (or 4.5.7) introduced a different utf16<->utf8 conversion (using `wstring_convert` instead of boost's conversion library), which considerably slows down the reading of string data - I observed 10x slowdown with 1m rows of characters. 

Apart from the performance improvement, the changes are warranted because [wstring_convert](https://en.cppreference.com/w/cpp/locale/wstring_convert) was deprecated in C++17.

## Changes

- Migrate away from `wstring_convert` in favor of [simdutf](https://github.com/simdutf/simdutf)
  - similar performance to the old `boost` implementation (no formal benchmark in this PR - just local tests),
  - is easily buildable and conda-packaged: similar setup with ninja and cmake, see [recipe](https://github.com/conda-forge/simdutf-feedstock/blob/main/recipe/build.sh)
  - slightly more cumbersome to use - very open to any feedback and comments
- Include Windows CI


cpp/turbodbc_arrow/Library/src/set_arrow_parameters.cpp
```diff
@@ -138,7 +139,7 @@ namespace {
         size_t maximum_length = 0;
         for (int64_t i = 0; i != elements; ++i) {
           if (!typed_array.IsNull(start + i)) {
-	    std::u16string str = std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}.from_bytes(typed_array.GetString(start + i));
+            std::u16string str = boost::locale::conv::utf_to_utf<char16_t>(typed_array.GetString(start + i));
             maximum_length = std::max(maximum_length, str.length());
```
SimeonStoykovQC, line 143: TODO: this change should be reverted & implemented with simdutf eventually - it was just to isolate the line that was causing the slowdown with my benchmarks. It's not this one.

xhochy:
Please add a changelog entry.

CHANGELOG.rst
```diff
@@ -3,6 +3,11 @@ Version history / changelog
 
 From version 2.0.0, turbodbc adapts semantic versioning.
 
+Version 4.9.1
```
xhochy, line 6: As we introduce a new dependency, I would bump the minor.
```suggestion
Version 4.10.0
```

PR #401, opened 2023-11-16
Add pyproject.toml to build artifact to meet pep517

cshaley:
pyproject.toml was missing in the tar.gz build artifact.  This caused the command `pip wheel --use-pep517 "turbodbc==4.8.0"` to fail.

I verified the problem with the below commands and then tested the change the same way.
```
cmake -DBOOST_ROOT=$CONDA_PREFIX -DBUILD_COVERAGE=ON -DCMAKE_INSTALL_PREFIX=./dist -DPYTHON_EXECUTABLE=$(which python) -GNinja .. && ninja && cmake --build . --target install && cd dist && python setup.py sdist

pip wheel --use-pep517 "dist/turbodbc-4.9.0.tar.gz"
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,11 +3,15 @@ Version history / changelog
 
 From version 2.0.0, turbodbc adapts semantic versioning.
 
+Version 4.9.0
+-------------
+
+* Fix pip wheel missing pyproject.toml - support pep517
+
 Version 4.8.0
 -------------
 
 * Support ``pyarrow=14``
-
 
 Version 4.7.0
 -------------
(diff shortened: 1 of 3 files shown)


rxm7706:
@cshaley @xhochy thank you !
I really wonder how many people care about PiP and conda packages releases being available to everyone as soon as possible - especially when a CVE is found and patched by upstream maintainers so quickly. 

But I'm glad to see how fast the Pyarrow contributors reacted, and we are close to having both TurboODBC and ML-Flow patched and available for use through PiP, Poetry and Conda. Thank you for your time,

PR #423, opened 2024-04-26
Update pyproject.toml to include pyarrow < 16

gblivegblive:
setup.py requires pyarrow<16, so should be reflected here as well


gblivegblive:
not sure why the tests are failing - it complains about missing MS ODBC driver.

xhochy:
> not sure why the tests are failing - it complains about missing MS ODBC driver.

Yes, we regularly need to bump the config when MS does a new release. I'll push a fix.

PR #426, opened 2024-05-13
Build and test with pyarrow 16

xhochy:


alippai:
This looks good, thanks!
No new release will be needed for 16.1.0, right?

xhochy:
No, that should also work with this PR

PR #433, opened 2024-08-26
Move to `meson-python`

xhochy:
Switch the build backend to `mesonpy`. This should future-proof this package and removes the need for two different build systems.

At the same time we're switching to `pixi` instead of using `conda` and Earthly together. We could probably utilise `pixi` tasks a bit more, but that is something for a future cleanup.


xhochy:
Currently waiting for https://github.com/conda-forge/psqlodbc-feedstock/pull/6

alippai:
This looks to be a great cleanup and improvement!

xhochy:
I'm really excited how straight-forward meson(-python) was to use.

xhochy:
@alippai @SimeonStoykovQC If one of you could do a review of this PR, I would highly appreciate it.

cpp/turbodbc_python/Library/src/python_bindings/module.cpp
```diff
@@ -1,4 +1,16 @@
+#include <turbodbc_arrow/arrow_result_set.h>
+#include <turbodbc_arrow/set_arrow_parameters.h>
+#include <turbodbc_numpy/numpy_result_set.h>
+#include <turbodbc_numpy/set_numpy_parameters.h>
+#include <turbodbc/cursor.h>
+
+// compare http://docs.scipy.org/doc/numpy/reference/c-api.array.html#importing-the-api
```
alippai, line 7: This is a broken link now

meson.build
```diff
@@ -0,0 +1,172 @@
...
+    check: true
+  ).stdout().strip()
+else
+  _incdir_numpy_abs = incdir_numpy
+endif
+inc_np = include_directories(incdir_numpy)
+# Don't use the deprecated NumPy C API. Define this to a fixed version instead of
+# NPY_API_VERSION in order not to break compilation for released SciPy versions
+# when NumPy introduces a new deprecation.
+numpy_nodepr_api = ['-DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION']
```
alippai, line 78: I see this is copied from SciPy, but I wonder why NPY_1_9_API_VERSION (or NPY_1_7_API_VERSION) and not the minimum numpy?

pixi.toml
```diff
@@ -0,0 +1,67 @@
...
+gtest = ">=1.15.0,<1.16"
+gmock = ">=1.15.0,<1.16"
+meson-python = ">=0.16.0,<0.17"
+ninja = ">=1.12.1,<1.13"
+c-compiler = "*"
+cxx-compiler = "*"
+numpy = ">=1.21,<3"
+cmake = ">=3.30.2,<3.31"
+libboost-devel = ">=1.86.0,<1.87"
+simdutf = ">=5.4.15,<5.5"
```
alippai, line 21: I don't think this matches the conda feedstock meta.yaml

alippai:
I have no C++ build experience, but looks much simpler than it was before.

xhochy:
Running into https://github.com/prefix-dev/setup-pixi/issues/139

Issue #443, opened 2024-10-16
turbodbc_arrow_support can’t be imported


alippai:
I get such error when using a pyarrow table as param to executemanycolumns() with the latest version. (Cc @xhochy)

Should it be in the arrow_support.py:
from turbodbc_intern import *

PR #444, opened 2024-10-18
Fix turbodbc_intern import and test pyarrow as a hard dependency

alippai:
Fix https://github.com/blue-yonder/turbodbc/issues/443


xhochy (approved):

xhochy:
Thanks!

PR #457, opened 2025-01-18
Support pyarrow 19

alippai:
@xhochy This just adds pyarrow 19, but I'm not sure about the build deps like meson and boost


xhochy:
We can update  them in separate PRs. Thanks!

xhochy (approved):

PR #462, opened 2025-02-12
Timeout

boekeljasper:
This adds a timeout argument to the "execute" method.


xhochy:
This looks very wrong. Please have a look at the diff.

pixi.toml
```diff
@@ -1,5 +1,5 @@
 [project]
-name = "turbodbc"
+name = "cnvxaturbodbc"
```
xhochy, line 2: What is this rename about?

PR #465, opened 2025-02-25
Add missing boost dependency to meson.build

alippai:


alippai:
@xhochy `boost` seems to be missing from the meson.build files, but I can't figure out how to make it detectable. Do you have any ideas?

alippai:
Alternatively I replaced boost with C++20: https://github.com/blue-yonder/turbodbc/pull/466

Issue #402, opened 2023-11-30
Does turbodbc still need boost as a dependency?


bgemmill:
The C++ standard library has been absorbing concepts from boost, and it looks like it may be close to providing everything turbodbc needs, depending on which standard turbodbc wants to target.

From [here](https://turbodbc.readthedocs.io/en/latest/pages/getting_started.html), it looks like the required boost libraries are variant, optional, datetime, and locale.

C++17 has [variant](https://en.cppreference.com/w/cpp/utility/variant), [optional](https://en.cppreference.com/w/cpp/utility/optional), C++11 has [chrono](https://en.cppreference.com/w/cpp/chrono) and C++20's chrono has a standardized [date](https://github.com/HowardHinnant/date), and it looks like boost::locale [may have been removed already](https://github.com/blue-yonder/turbodbc/commit/882fb2f6614c0fa0a3a2b25a4b3339dc6e2d819f).

Depending on the date formatting required, it may be too soon to consider requiring the standard library from 20, but this seemed interesting to bring up.

alippai:
Does this look good? https://github.com/blue-yonder/turbodbc/pull/466

PR #466, opened 2025-02-25
Remove boost

alippai:


cpp/turbodbc/Library/src/parameter_sets/set_field.cpp
```diff
@@ -65,36 +66,41 @@ namespace {
...
+			auto hrs   = std::chrono::duration_cast<std::chrono::hours>(time_of_day);
+			time_of_day -= hrs;
+			auto mins  = std::chrono::duration_cast<std::chrono::minutes>(time_of_day);
+			time_of_day -= mins;
+			auto secs  = std::chrono::duration_cast<std::chrono::seconds>(time_of_day);
+			time_of_day -= secs;
+			auto micros = std::chrono::duration_cast<std::chrono::microseconds>(time_of_day);
 
+			auto destination = reinterpret_cast<SQL_TIMESTAMP_STRUCT *>(destination_.data_pointer);
+			destination->year = static_cast<SQLSMALLINT>(int(ymd.year()));
```
pacman82, line 88: No intention, or time to review the entire PR. However I can point out that:

`int ( ... )` is an old C style syntax for a static cast.
`static_cast<...>(...)` is the C++ native syntax for static cast.

So what is written here is a static cast, nested inside a static cast. If it is required likely only the outer one is needed.

Cheers, Markus

cpp/turbodbc/Library/src/time_helpers.cpp
```diff
@@ -5,80 +5,131 @@
...
-}
+// Alias a time_point with microsecond precision.
+using sys_time_us = std::chrono::time_point<std::chrono::system_clock, std::chrono::microseconds>;
 
+// Define the epoch for timestamps and dates.
+constexpr sys_time_us timestamp_epoch = 
+    sys_time_us(std::chrono::time_point_cast<std::chrono::microseconds>(
+        std::chrono::sys_days{std::chrono::year{1970}/1/1}
+    ));
+constexpr std::chrono::sys_days date_epoch = std::chrono::sys_days{std::chrono::year{1970}/1/1};
```
bgemmill, line 20: Unless there's another reason to manually encode the epoch, the default constructor of time_point [does this for you](https://en.cppreference.com/w/cpp/chrono/time_point/time_point).

cpp/turbodbc/Library/src/time_helpers.cpp
```diff
@@ -5,80 +5,131 @@
...
+    std::chrono::year_month_day ymd{
+        std::chrono::year{sql_ts.year},
+        std::chrono::month{sql_ts.month},
+        std::chrono::day{sql_ts.day}
+    };
+    std::chrono::sys_days date_part{ymd};
+
+    // Build the time-of-day portion.
+    // Note: sql_ts.fraction is assumed to be in nanoseconds.
+    auto micros_from_fraction = std::chrono::microseconds(sql_ts.fraction / 1000);
```
bgemmill, line 36: [std::chrono::nanoseconds](https://en.cppreference.com/w/cpp/chrono/duration) seems like a better fit given the input type, and this is time_point_cast<> to microseconds below anyway.

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -422,21 +430,33 @@ TEST_F(ArrowResultSetTest, MultiBatchConversionTimestamp)
                 }
             } else {
                 ASSERT_OK(builder.Append(i));
-                auto td = boost::posix_time::microseconds(i);
-                auto ts = boost::posix_time::ptime(boost::gregorian::date(1970, 1, 1), td);
+                auto tp = std::chrono::system_clock::time_point{
+                    std::chrono::sys_days{std::chrono::year{1970}/std::chrono::month{1}/std::chrono::day{1}} +
```
bgemmill, line 434: Default constructing a time point and adding one microsecond prevents manually encoding the epoch.

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
```diff
@@ -451,21 +471,33 @@ TEST_F(ArrowResultSetTest, MultiBatchConversionTimestamp)
         arrow::TimestampBuilder builder(arrow::timestamp(arrow::TimeUnit::MICRO), pool);
         for (int64_t i = 0; i < 2 * OUTPUT_SIZE; i++) {
             ASSERT_OK(builder.Append(i));
-            auto td = boost::posix_time::microseconds(i);
-            auto ts = boost::posix_time::ptime(boost::gregorian::date(1970, 1, 1), td);
+            auto tp = std::chrono::system_clock::time_point{
+                    std::chrono::sys_days{std::chrono::year{1970}/std::chrono::month{1}/std::chrono::day{1}} +
```
bgemmill, line 475: Same here too :-)

meson.build
```diff
@@ -4,7 +4,7 @@ project(
   meson_version: '>= 1.1.0',
   default_options: [
     'buildtype=release',
-    'cpp_std=c++17',
+    'cpp_std=c++20',
```
bgemmill, line 7: I didn't see any of cpp 20's date formatting, this may all just work in 17.

bgemmill (commented):
Generally looks good to me, I had a few nits around nanosecond handling and epoch encocing, and it looks like all of this may just require cpp17 rather than 20.

Thanks for looping me in from the [issue](https://github.com/blue-yonder/turbodbc/issues/402)!

cpp/turbodbc/Library/src/parameter_sets/set_field.cpp
alippai: Thanks for having a look at it!
Looks like the nested casts were needed, with the outer cast only it didn't compile. Made the syntax consistent.

cpp/turbodbc/Library/src/time_helpers.cpp
alippai: Thanks, updated!

cpp/turbodbc/Library/src/time_helpers.cpp
alippai: Updated

cpp/turbodbc/Library/src/time_helpers.cpp
```diff
@@ -5,80 +5,131 @@
...
+    auto time_of_day = tp - dp;
+    auto hrs   = std::chrono::duration_cast<std::chrono::hours>(time_of_day);
+    time_of_day -= hrs;
+    auto mins  = std::chrono::duration_cast<std::chrono::minutes>(time_of_day);
+    time_of_day -= mins;
+    auto secs  = std::chrono::duration_cast<std::chrono::seconds>(time_of_day);
+    time_of_day -= secs;
+    auto micros = std::chrono::duration_cast<std::chrono::microseconds>(time_of_day);
+
+    auto remainder_nanoseconds = nanoseconds % 1000;
```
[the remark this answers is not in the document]
alippai: Updated

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
alippai: Updated

cpp/turbodbc_arrow/Test/tests/arrow_result_set_test.cpp
alippai: Updated

meson.build
alippai: It didn't compile with c++17 for me. I'm not familiar with C++ ecosystem, is the newer version problematic eg. in the python builds?

alippai:
@bgemmill thanks for the help. I did my best to address the comments.

meson.build
bgemmill: If the system clang or gcc is new enough to handle cpp20 it shouldn't be an issue building; it's more a question of how old of a toolchain does turbodbc want to support. Since I'm not a maintainer, I'm likely not the best guy to answer that one ;-)

alippai: C++17 on this PR gives:
```
../cpp/turbodbc/Library/turbodbc/field.h:14:28: error: 'year_month_day' in namespace 'std::chrono' does not name a type
   14 | using date  = std::chrono::year_month_day;```

alippai: Turbodbc mostly follows the SPEC-0 and NEP-29 which means python 3.11 support.
The OS versions shipping with python 3.11 also have gcc 11 which supports C++20 chrono. 
The OS versions in the CI matrix also passed, so this should be good (also it's not something hard to revert partially or completely).

xhochy:
Thanks! This looks really promising. Could you try and do some kind of performance test? The last time I removed some boost code the performance got worse and we needed to introduce `simdutf` as an additional dependency.

